PR #160, opened 2015-08-06
UMVE: Small UI fixes

keksberg:


keksberg:
The window placement now seems to work properly.

PR #182, opened 2015-09-20
Cmake

Smozeley:


simonfuhrmann:
These Windows related commits will not be part of MVE and remain in Andre's repository.

PR #192, opened 2015-11-11
UTIL: Change FileLock to meet the BasicLockable requirements

magcks:
BasicLockable: http://en.cppreference.com/w/cpp/concept/BasicLockable

Example with std::lock_guard:
```c++
#include "util/file_system.h"
#include <mutex>
#include <thread>
#include <chrono>
#include <iostream>

int main (int argc, char** argv)
{
    util::fs::FileLock fl("test.png");
    {
        std::lock_guard<util::fs::FileLock> lock(fl); // lock()
        std::this_thread::sleep_for(std::chrono::seconds(2));
        // Implicit unlock()
    }
    std::cout << "Is locked: " << fl.is_locked() << std::endl; // No!
}
```

### DIFF
--- a/libs/util/file_system.cc
+++ b/libs/util/file_system.cc
@@ -15,6 +15,8 @@
 #include <cstring> // std::strerror
 #include <cstdio> // std::rename
 #include <algorithm>
+#include <thread>
+#include <chrono>
 
 #if defined(_WIN32)
 #   include <direct.h>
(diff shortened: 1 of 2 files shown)


magcks:
Note: The util::FileLock class was unused. The change of the API shouldn't break anything.

magcks:
There are concepts for allocators, too. Does this mean, that we shouldn't implement our own allocators anymore? The cppreference site of std::vector says "The type must meet the requirements of Allocator." and for std::lock_guard the type must meet the "BasicLockable" "requirements". I think they have written "requirements" instead of "concept" on purpose.

http://en.cppreference.com/w/cpp/concept/Allocator
http://en.cppreference.com/w/cpp/container/vector
http://en.cppreference.com/w/cpp/thread/lock_guard

libs/util/file_system.cc
```diff
@@ -586,92 +588,61 @@ Directory::scan (std::string const& path)
...
 }
 
 bool
-FileLock::release (void)
+FileLock::lockable (void)
 {
-    if (this->lockfile.empty())
-        return false;
-    return fs::unlink(this->lockfile.c_str());
+    return !!this->filename;
```
keksberg: Better write
```c++
return this->filename != nullptr;
```
here.

simonfuhrmann:
Since I don't see a good reason why we need this now, I'm closing this.

PR #193, opened 2015-11-11
UTIL: Replace AlignedMemory with std::vector and AlignedAllocator

magcks:


magcks:
Processed your annotations.

PR #202, opened 2016-01-21
Implement feature matching with cascade hashing

andre-schulz:
Using cascade hashing for feature matching trades a slight drop in matching
accuracy for a speed-up of an order of magnitude
Speed-up over bruteforce matching on the citywall-20140923 dataset:
1x Intel Core i7-3930K   : ~20.6x
2x Intel Xeon E5-2687W v3: ~18.1x

### DIFF
--- a/apps/sfmrecon/sfmrecon.cc
+++ b/apps/sfmrecon/sfmrecon.cc
@@ -58,6 +58,7 @@ struct AppSettings
     int initial_pair_1 = -1;
     int initial_pair_2 = -1;
     int min_views_per_track = 3;
+    bool cascade_hashing = false;
 };
 
 void
@@ -111,6 +112,9 @@ features_and_matching (mve::Scene::Ptr scene, AppSettings const& conf,
     matching_opts.ransac_opts.verbose_output = false;
     matching_opts.use_lowres_matching = conf.lowres_matching;
     matching_opts.match_num_previous_frames = conf.video_matching;
+    matching_opts.matcher_type = conf.cascade_hashing
+        ? sfm::bundler::Matching::MATCHER_CASCADE_HASHING
+        : sfm::bundler::Matching::MATCHER_EXHAUSTIVE;
 
     std::cout << "Performing feature matching..." << std::endl;
     {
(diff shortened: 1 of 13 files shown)


andre-schulz:
Updated

simonfuhrmann:
Added comments.

simonfuhrmann:
Answers in my original comments because of some Github crazyness.

andre-schulz:
I chose to keep the commit that changes the definition of [`{Sift,Surf}Descriptors`](https://github.com/simonfuhrmann/mve/pull/202/commits/2e393073b4e393d6a1767111e2c20b98a024e4f4#diff-e2119bdb694f19daff51c7c6ce69214dL47) from `util::AlignedMemory<{{u}int16_t,float}>` to `util::AlignedMemory<math::Vec{128{f,us},64{f,s}}` because of the following reasons in addition to being more expressive:
- Allows one to call `{Sift,Surf}Descriptors::size()` and get the actual number of descriptors and it also makes the [counters](https://github.com/simonfuhrmann/mve/pull/202/commits/2e393073b4e393d6a1767111e2c20b98a024e4f4#diff-e2119bdb694f19daff51c7c6ce69214dL61) in `ProcessedFeatureSet` superfluous
- [`CascadeHashing::collect_features_from_buckets()`](https://github.com/simonfuhrmann/mve/pull/202/commits/216263f2679740ca9845f551ae0ba8c76ea2880b#diff-04d556e206d55dfcfa720b5ef26ad6edR425) is ~12% faster than a non-templatized version.

KevinCain:
I've been excited to follow your implementation of cascade hashing for MVE feature matching, congratulations on this latest work.

In my trials with MVE, feature matching accounts for a sizable percentage of total running time.  Bundle adjustment, however, I've found an even more pressing issue for larger data sets (e.g.:  several thousand large images).  This is especially true when running on a virtual cluster, as bundle adjustment is not yet ready for parallel computation.

Does your new feature search code suggest a path for parallel bundle adjustment?

simonfuhrmann:
No, unfortunately parallel bundle adjustment is a difficult problem. As you might have seen, we implemented our own bundle adjustment code (in the imBA branch) which doesn't really make use of any parallelization, and although it's a bit slower than PBA, the difference is not significant.

PR #205, opened 2016-01-22
Generalize percentile implementation

nmoehrle:
Since this could be useful for different applications I think it should eventually be moved to *libs/math/algo.h*.

### DIFF
--- a/apps/meshclean/meshclean.cc
+++ b/apps/meshclean/meshclean.cc
@@ -33,15 +33,40 @@ struct AppSettings
     int component_size;
 };
 
-template <typename T>
-T
-percentile (std::vector<T> const& in, float percent)
-{
-    /* copy the input vector because 'nth_element' will rearrange it */
-    std::vector<T> copy = in;
-    std::size_t n = static_cast<std::size_t>(percent / 100.0f * in.size());
-    std::nth_element(copy.begin(), copy.begin() + n, copy.end());
-    return copy[n];
+
+template <class InputIt>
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
I don't get it. Did you re-implement `std::nth_element`? Why?

nmoehrle:
Because I don't want to duplicate the values vector (```std::nth_element``` is modifying it) and computing the percentile is possible with a single iteration over the values.

simonfuhrmann:
Reasons for rejection:
- Generally, we don't need this now.
- Specifically, we don't need this in `meshclean` (should have been submitted to `libs/math/algo.h`)
- No tests
- Does not work with unsigned values (because `invert`)
- Also copies up to 50% of the values, plus expensive heap operations (performance probably worse than copying and using `std::nth_element`)

nmoehrle:
Worse case performance on normally distributed input values, wrt. runtime, is indeed six times worse than copying and and using ```std::nth_element```.

The implementation works for unsigned types as an result of how integer conversion is defined in the cpp standard (Section 4.7.2):
> If the destination type is unsigned, the resulting value is the least unsigned integer congruent to the source integer (modulo 2 n where n is the number of bits used to represent the unsigned type).
[Note: In a two’s complement representation, this conversion is conceptual and there is no change in the bit pattern (if there is no truncation). — end note ]

So if ```a < b``` then ```-a > -b``` holds for signed and unsigned types.

PR #207, opened 2016-02-05
Refactor matching code

andre-schulz:
- Move matching code from `FeatureSet` into its own class(es)
- Introduce `MatchingBase` class which consists of data structures (`Options`, `Result`), helper methods and an interface which are common to different matchers.
- Implement `ExhaustiveMatching` as the first matcher
- Add a pointer to `MatchingBase` in `sfm::bundler::Matching` to support different matchers in the future


simonfuhrmann:
As discussed, `sfm/matching.h` should stay and be used from `ExhaustiveMatching` and potentially other places.

simonfuhrmann:
Almost there.

PR #212, opened 2016-03-18
SFM: Fix compilation of cascade hash matcher on Visual Studio

andre-schulz:
- popcntintrin.h is not provided by Visual Studio 2015 but nmmintrin.h
  includes _mm_popcnt_u64()


andre-schulz:
Updated

caomw:
good

PR #217, opened 2016-03-30
Refactor matrix/vector dimension(s)

andre-schulz:
- Querying the dimension of a vector or the number of rows/columns of a
  matrix is now done by accessing (a) static const variable(s) instead of
  calling a method.
- The refactoring of the vector class also works around a compiler bug related
  to the constexpr keyword in Visual Studio 2015 Update 1.

### DIFF
--- a/tests/math/gtest_matrix.cc
+++ b/tests/math/gtest_matrix.cc
@@ -80,6 +80,8 @@ TEST(MatrixTest, MatrixOperations)
     test(2,0) = 7.0f; test(2,1) = 8.0f; test(2,2) = 9.0f;
 
     /* Matrix access, min, max, square check, . */
+    EXPECT_EQ((Matrix<int,3,4>::rows), 3);
+    EXPECT_EQ((Matrix<int,3,4>::cols), 4);
     EXPECT_EQ(test.col(1), Vec3f(2.0f, 5.0f, 8.0f));
     EXPECT_EQ(test.row(1), Vec3f(4.0f, 5.0f, 6.0f));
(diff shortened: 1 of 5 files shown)


tests/math/gtest_matrix.cc
```diff
@@ -80,6 +80,8 @@ TEST(MatrixTest, MatrixOperations)
     test(2,0) = 7.0f; test(2,1) = 8.0f; test(2,2) = 9.0f;
 
     /* Matrix access, min, max, square check, . */
+    EXPECT_EQ((Matrix<int,3,4>()).rows, 3);
```
simonfuhrmann: What about `Matrix<int,3,4>::rows`?

tests/math/gtest_vector.cc
```diff
@@ -16,9 +16,9 @@ TEST(VectorTest, MiscOperations)
     EXPECT_EQ(Vec2f(10.0f, 0.0f).normalize(), Vec2f(1.0f, 0.0f));
     EXPECT_EQ(Vec2f(3.0f, 4.0f).norm(), 5.0f);
     EXPECT_EQ(Vec2f(2.0f, 2.0f).square_norm(), 8.0f);
-    EXPECT_EQ(Vec2f().dim(), 2);
-    EXPECT_EQ(Vec3f().dim(), 3);
-    EXPECT_EQ(Vec4f().dim(), 4);
+    EXPECT_EQ(Vec2f().dim, 2);
```
simonfuhrmann: `Vec2f::dim`

PR #222, opened 2016-04-21
Output of Visitibility Information & High Resolution Photosynther Images Import

SamirAroudj:
Additional visibility information output:
- dmrecon: output of images with view IDs from local view selection for reconstructed depth values if wanted
- ply_mesh
- scene2pset: output of view IDs as vertex element properties if wanted
- mesh_io_ply.cc: support of view ID vertex element properties for loading
- umve: hide view ID images as it doesn't make sense to show them
- scene.cc: if scene is saved cameras.txt file is created (contains positions etc. of all cameras)

Changes for photosynther:
- makescene: usage of images from l0 folder to import high resolution undistorted images into the scene instead of downscaled for_stereo... images if available

### DIFF
--- a/apps/dmrecon/dmrecon.cc
+++ b/apps/dmrecon/dmrecon.cc
@@ -155,6 +155,8 @@ main (int argc, char** argv)
         "progress output style: 'silent', 'simple' or 'fancy'");
     args.add_option('\0', "force", false,
         "Reconstruct and overwrite existing depthmaps");
+	args.add_option('V', "keep-views", false,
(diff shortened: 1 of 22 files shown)


flanggut:
Aehm, this is srsly fucked up IMHO. Indentation is simply broken. And in general - you cannot make random API changes just to accommodate your needs.

simonfuhrmann:
Please provide a single PR for each feature. However, I will not pull "additional visibility information output", as it's simply not useful for the general public.

PR #226, opened 2016-04-29
Sanitize path when fetching basename to make it Windows-compatible

igorti:
`basename` function fails on Windows as it only takes into consideration forward slashes. This can easily be fixed by sanitizing the path first.

### DIFF
--- a/libs/util/file_system.cc
+++ b/libs/util/file_system.cc
@@ -426,19 +426,21 @@ dirname (std::string const& path)
 std::string
 basename (std::string const& path)
 {
+    std::string sanitized = sanitize_path(path)
+    
     /* Skip group of slashes at the end. */
-    std::size_t len = path.size();
-    while (len > 0 && path[len - 1] == '/')
+    std::size_t len = sanitized.size();
+    while (len > 0 && sanitized[len - 1] == '/')
         len -= 1;
     if (len == 0)
         return "";
 
     /* Skip basename. */
     std::size_t base = len - 1;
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
This PR changes the behavior of basename. For example. for `a/b/..` you would expect `..` as result, however, now you would get `a`. All functions here work with slashes and, in fact, no backslashes should ever be present in paths. Paths that come from user inputs, such as arguments from apps, should be sanitized right away. Are you experiencing any issues with our tools on Windows?

simonfuhrmann:
Any comments to my concerns? Where are you experiencing issues?

igorti:
Sorry for not getting back, I didn't have time to verify your concerns. The problem I stumbled upon was in mvs-texturing. When running it on Windows it was writing an absolute path for to the `.mtl` file (should only be the file name) when failing [to fetch basename here](https://github.com/nmoehrle/mvs-texturing/blob/dab68acaa693275c183c254a958130ee6d29c3e4/libs/tex/obj_model.cpp#L32). 

So, do you propose that sanitizing should be done prior to calling `basename` function? Isn't it better to put this logic into the function itself so that it works on any platform and you don't have to bother about what kind of input it expects?

simonfuhrmann:
Sanitizing should be done on the user input once. That's a simple rule, and IMHO better than putting it in every function that deals with paths, especially because it can change the behavior of functions (see the example I made). 

I will ping the MVS texturing guys to fix it. 

Thanks.

PR #240, opened 2016-06-19
Add loading of mvei images to load_image - fail fast version

nmoehrle:
- avoids loading image content of wrong type
- loads header twice for correct type

### DIFF
--- a/libs/mve/image_io.cc
+++ b/libs/mve/image_io.cc
@@ -83,6 +83,17 @@ load_file (std::string const& filename)
         { return load_ppm_file(filename); }
         catch (util::FileException& e) { throw; }
         catch (util::Exception& e) {}
+
+        try
+        {
+            ImageHeaders header = load_mvei_file_headers(filename);
+            if (header.type != IMAGE_TYPE_UINT8)
+                throw util::Exception("Invalid image format");
+            ImageBase::Ptr image = load_mvei_file(filename);
+            return std::dynamic_pointer_cast<ByteImage>(image);
+        }
+        catch (util::FileException& e) { throw; }
+        catch (util::Exception& e) {}
     }
     catch (util::FileException& e)
     {


libs/mve/image_io.cc
```diff
@@ -87,7 +87,9 @@ load_file (std::string const& filename)
         try
         {
             ImageHeaders header = load_mvei_file_headers(filename);
-            if (header.type != IMAGE_TYPE_UINT8) { throw; }
+            if (header.type != IMAGE_TYPE_UINT8) {
```
simonfuhrmann: Please remove brackets here (1. not according to style, 2. not necessary)

libs/mve/image_io.cc
```diff
@@ -87,7 +87,9 @@ load_file (std::string const& filename)
...
+                throw util::Exception("Invalid image format");
+            }
             ImageBase::Ptr image = load_mvei_file(filename);
             return std::static_pointer_cast<ByteImage>(image);
```
simonfuhrmann: I belive this should be a `std::dynamic_pointer_cast`

simonfuhrmann:
I think I prefer the fail-fast version.
Can you please force-push with only one commit prefixed with `MVE: `. Thanks!

libs/mve/image_io.cc
nmoehrle: Because the type is essentially checked manually before I thought it is unnecessary do another runtime check with `std::dynamic_pointer_cast`. However, this assumes that the file does not change between the `load_mvei_file_headers` and `load_mvei_file`. I guess the runtime check is preferable in this case.

simonfuhrmann:
Thanks. In any case, doing a dynamic cast ensures that you don't re-interpret the pointer type. Because that lead's to segfault.

PR #243, opened 2016-06-23
Fix descriptor discretization

nmoehrle:
An error within the discretization limited the range from 65536 to 256 values


nmoehrle:
Apparently this effect is desired

PR #244, opened 2016-06-23
SFM: Add surf features to matching test app

nmoehrle:
Further these commits:
- eliminate code duplication (`Matching::combine_results`)
- fix an uninitialized value bug (`sfm::Matching::Options::distance_threshold`)
- allow the inspection of  discretized matching results (which were found to be quite different)

### DIFF
--- a/libs/sfm/_test_matching.cc
+++ b/libs/sfm/_test_matching.cc
@@ -47,16 +47,52 @@ visualize_matching (sfm::Matching::Result const& matching,
     return match_image;
 }
 
+#define DISCRETIZE_DESCRIPTORS 0
+template <typename T>
 void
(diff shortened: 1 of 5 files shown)


simonfuhrmann:
Where is the unintialized value bug?

nmoehrle:
Just in the matching test app, `sfm::Matching::Options::distance_threshold` is not initialized by a constructor and thus has to be initialized manually. This [comment](https://github.com/simonfuhrmann/mve/blob/master/libs/sfm/matching.h#L45) is also a misleading in that sense.

I though about adding a constructor and forcing the initialization, but I know that you think, and I agree, that struct initialization is more readable by member access. Adding a default constructor is problematic because there is no right value for `sfm::Matching::Options::descriptor_length`. I haven't been able to come up with a good solution.

PR #253, opened 2016-08-24
add a multi thread for makescene app

xialang2012:


simonfuhrmann:
Adding multi-threading to this function is non-trivial. It has outside-of-loop counters, and will produce race conditions if parallelized like this. Sorry. We can look into it if you re-open the feature request.

pierotofy:
Sorry to comment on such an old thread.

I'm also looking to add multi-threading to `makescene`. I've looked at the inner loop of this PR (starting at line 560) and I can't seem to find obvious places where racing conditions would occur (but I'm not too familiar with the mve codebase). 

To do some testing for potential side effects I've compiled two versions of `makescene`, one with the original (single threaded) loop and one with the changes in this PR (multi threaded). I've run the same program with a Noah's Bundler's format input with both, running the multi threaded version multiple times and then diffed the folders with:

`diff -r output_single_thread/ output_multi_thread/`

Between different runs the `diff` command found no differences in output. I know this is not enough to cover all bases and there could be problems elsewhere, but I'm hoping somebody could point out the trouble spots where I should pay attention for race conditions.

simonfuhrmann:
Well, see how this function, or the loop, uses the variable `num_valid_cams` and increments the variable in a different place? That leads to a race condition.

I guess this is only relevant if you're using Photosynther, and it should work Noah's bundler, but that is the reason this PR was initially closed. Also note that the instruction `num_valid_cams++` can lead to a race condition, because this kind of increment is not atomic (i.e., variable read, increment, write).

On another note, why are you using Noah's bundler? There are more modern, more efficient solutions out there these days. VisualSfM, OpenMVG and even MVS's own SfM code.

pierotofy:
Hey @simonfuhrmann thanks so much for the reply! :pray: 

Thank you for pointing that out. It makes sense. I have opened #460 which should address the concerns.

Ah, that's a good question, in OpenDroneMap (https://github.com/OpenDroneMap/ODM) we actually use OpenSfM (https://github.com/mapillary/OpenSfM), but to create an MVE scene we use Noah's bundler format as an intermediary format (opensfm has a Noah's bundler export) and then we import it into MVE.

opensfm --> noah --> MVE

We ideally would want to export directly to MVE, but we need to find some time to write an exporter in opensfm or a makescene importer. It's on the list of things to do :)

Issue #256, opened 2016-09-26
Gamma correction sRGB


nmoehrle:
The sRGB gamma correction within MVE is based on the sRGB proposal found at http://www.w3.org/Graphics/Color/sRGB rather than the standard IEC 61966-2-1.

A note on the site itself marks it as obsolete and mentions a change of constants:
> Note: This document is obsolete, and is retained here for historical purposes only. It was published on 5 November 1996, as a proposal specification for sRGB as a standard default color space. sRGB has since been standardized within the International Electrotechnical Commission (IEC) as IEC 61966-2-1. During standardization, a small numerical error caused by rounding error was corrected. The viewing conditions were also clarified.

The standard itself is not openly accessible, however, the linear ramp thresholds commonly referred to as  standard differ from the ones originally proposed. An explanation for the different values can be found [here](https://en.wikipedia.org/wiki/SRGB#Theory_of_the_transformation).

Thresholds for linear ramp

 | linear -> sRGB | sRGB -> linear
-------------|:-----------:|:------------:
Proposal | 0.00304 | 0.03928
Standard | 0.00313 | 0.04045

Further the handling of gamma encoded values within MVE in general should be discussed.
All input images are presumably sRGB encoded with non-linear light intensities. Within our pipeline these values are then rescaled, undistorted(sfmrecon), blended(fssrrecon) and shaded(umve) without ever being decoded.

A well written article (by John Novak) on gamma correction and the effects of gamma-incorrectness can be found [here](http://blog.johnnovak.net/2016/09/21/what-every-coder-should-know-about-gamma/).

simonfuhrmann:
Feel free to send a PR.

simonfuhrmann:
Regarding processing of already gamma-corrected values: While it is true that these values should be linearized before processing them, I don't think that there is a reasonable solution for it.

1. We don't know how these images have been converted from linear intensities to RGB images. Every camera applies a different, unknown CRF and images might be post-processed (Lightroom/Photoshop).
2. Conversion before and after every operation increases computation and makes the code less readable.
3. Blending in FSSR was never meant to produce useful textures, it does not even consider different exposure.

I'm not sure if we gain anything for handling these things "properly", and I don't think it's possible to handle it properly.

nmoehrle:
Opened PR #257

I agree that the framework should remain general but my assumption is that most of the images are sRGB encoded and the images that are not have been taken by someone who would be able to encode them.
[...]

PR #257, opened 2016-09-26
Use sRGB linear ramp threshold of implemented standard #256

nmoehrle:


nmoehrle:
We have a test for sRGB gamma correction [here](https://github.com/simonfuhrmann/mve/blob/master/tests/mve/gtest_image_tools.cc#L369). Should I add a test with some "gold" values or a tests that is checking the threshold and fails before this commit?

simonfuhrmann:
Yes, please add some golden values and epsilon-compare with them. It makes sure we don't break the implementation of the standard. The question is, where do we get reliable golden values from.

simonfuhrmann:
Updates?

nmoehrle:
I could not find any listing of gold values for sRGB conversions. Other library developers such as Nathaniel Smith have created their own set of [gold values](https://github.com/njsmith/colorspacious/blob/master/colorspacious/gold_values.py) based on the color conversion calculators of [http://www.brucelindbloom.com](http://www.brucelindbloom.com/index.html?ColorCalculator.html) and other sites.
So I guess we will have choose a trustworthy converter and generate some ourselves. I would personally vote for Nathaniels [colorspacious](https://github.com/njsmith/colorspacious).

simonfuhrmann:
SGTM

Issue #264, opened 2016-10-19
Memory consumption of initial pair search


nmoehrle:
The least squares fundamental estimation within the initial pair search becomes a memory bottleneck for scenes that have a large amount of pairwise matches. The cause is the size of the temporary matrices inside the SVD implementation.

For a scene with high resolution images and around 25000 pairwise matches this can lead to a memory consumption of whopping 33 GB:
```
->97.42% (35,549,996,368B) 0x447CB0: void math::internal::matrix_r_svd<double>(double const*, int, int, double*, double*, double*, double const&) (new_allocator.h:104)
| ->97.42% (35,549,996,368B) 0x441EBB: sfm::fundamental_least_squares(std::vector<sfm::Correspondence2D2D, std::allocator<sfm::Correspondence2D2D> > const&, math::Matrix<double, 3, 3>*) (matrix_svd.h:730)
|   ->97.42% (35,549,996,368B) 0x43448C: sfm::bundler::InitialPair::compute_pair(sfm::bundler::InitialPair::Result*) [clone ._omp_fn.0] (bundler_init_pair.cc:226)
```

simonfuhrmann:
It is indeed a problem. @flanggut, any ideas how to solve this?

 * We can limit the size of points we consider to a few hundreds. This should be enough.
 * Maybe we don't need this step and we rely on BA to fix the inaccuracies from not computing the consensus Fundamental matrix.

flanggut:
I think ideally the re-estimation from all matches ensures that we can triangulate as many matches as possible for the initial pair. This hasn't worked out super effective so far - I still see a lot of the successful initial pair searches run into problems in the initial triangulation step (especially in the context of CR-PLAY with not very ideal datasets). I actually implemented a fix to make the search more robust for the context of the project f1db54cd4f03f8bbccf9d90091716267cd5ae673. It simply checks if the triangulation for the initial pair actually works for a significant amount of the matches before it returns the pair.
Overall I'd suggest we do limit input of the estimation to a certain amount of (maybe randomly selected) points, but also check - similar to f1db54cd4f03f8bbccf9d90091716267cd5ae673 - if the triangulation still works for most of the points before we return the pair.

nmoehrle:
#265 limits the amount to 1000 random matches.

Checking the triangulation (f1db54c) solved an initial pair selection issue for me today. Most of the tracks from the automatically selected initial pair have been rejected by the subsequent triangulation which resulted in immediate termination.
```
[...]

PR #266, opened 2016-10-19
SFM: Use RANSAC instead of least squares approach for initial pair search. #264

stepankonrad:
This is a possible solution to Issue #264 and an alternative to #265. Using a RANSAC instead of a least squares fundamental matrix calculation maybe is overkill because the correspondences are already filtered. However, the memory consumption is very low. [In COLMAP, they also use a RANSAC for this step](https://github.com/colmap/colmap/blob/master/src/sfm/incremental_mapper.cc#L962).
The final solution can still be discussed...

### DIFF
--- a/libs/sfm/bundler_init_pair.cc
+++ b/libs/sfm/bundler_init_pair.cc
@@ -12,6 +12,7 @@
 #include <sstream>
 
 #include "sfm/ransac_homography.h"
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
Although I like this approach, using the fundamental matrix already computed in RANSAC would be better. Second, it might affect accuracy. Closing this due to the other approach we are using.

PR #270, opened 2016-10-25
SFM: Fix initial pair failure when triangulation fails

flanggut:
Add an additional check to the initial pair selection to test if two-view triangulation works for a significant amount of matches.

- This fixes many failures that usually occur when the two-view geometry from the initial pair selection is not optimal and a continued bundling is not possible.  
- Due to this we can also relax the homography test a little bit; which helps significantly for urban scenes where large homographies are very common but it is still possible to find an initial pair. 

We already tested this internally in the context of CR-PLAY.

### DIFF
--- a/libs/sfm/bundler_init_pair.cc
+++ b/libs/sfm/bundler_init_pair.cc
@@ -10,6 +10,7 @@
 #include <iostream>
 #include <vector>
 #include <sstream>
+#include <random>
 
 #include "sfm/ransac_homography.h"
 #include "sfm/triangulate.h"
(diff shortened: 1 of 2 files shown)


libs/sfm/bundler_init_pair.cc
```diff
@@ -84,6 +84,26 @@ InitialPair::compute_pair (Result* result)
...
             continue;
 
+        /* Run triangulation to ensure correct pair */
+        Triangulate::Options triangulate_opts;
+        Triangulate::Statistics stats;
+        Triangulate triangulator(triangulate_opts);
+        std::vector<CameraPose const*> poses;
+        poses.push_back(&pose1);
+        poses.push_back(&pose2);
+        std::size_t successfull_triangulations = 0;
```
simonfuhrmann: successful_triangulations

libs/sfm/bundler_init_pair.cc
```diff
@@ -84,6 +84,26 @@ InitialPair::compute_pair (Result* result)
...
+        for (std::size_t j = 0; j < candidate.matches.size(); ++j)
+        {
+            std::vector<math::Vec2f> positions;
+            positions.emplace_back(candidate.matches[j].p1);
+            positions.emplace_back(candidate.matches[j].p2);
+            math::Vec3d pos3d;
+            if(triangulator.triangulate(poses, positions, &pos3d, &stats))
```
simonfuhrmann: space after if

libs/sfm/bundler_init_pair.cc
```diff
@@ -84,6 +84,26 @@ InitialPair::compute_pair (Result* result)
         if (angle < this->opts.min_triangulation_angle)
             continue;
 
+        /* Run triangulation to ensure correct pair */
+        Triangulate::Options triangulate_opts;
+        Triangulate::Statistics stats;
```
simonfuhrmann: Move this line above the loop

libs/sfm/bundler_init_pair.cc
```diff
@@ -84,6 +84,26 @@ InitialPair::compute_pair (Result* result)
...
+        Triangulate triangulator(triangulate_opts);
+        std::vector<CameraPose const*> poses;
+        poses.push_back(&pose1);
+        poses.push_back(&pose2);
+        std::size_t successfull_triangulations = 0;
+        for (std::size_t j = 0; j < candidate.matches.size(); ++j)
+        {
+            std::vector<math::Vec2f> positions;
```
simonfuhrmann: I think it's more efficient if you keep this outside the loop and resize(2), then just assign in the loop.

flanggut:
Added changes.
NEW: Now also fixes clang compile error due to missing `<random>` include for `std::mt19937` from #265.

PR #272, opened 2016-10-26
APPS: read in cam files if available

nmoehrle:
This pull request would enable users to provide cam files along their image files to specify extrinsic and intrinsic camera parameters.

A cam file contains the camera part of the old MVE file header.

tx ty tz R00 R01 R02 R10 R11 R12 R20 R21 R22
f d0 d1 paspect ppx ppy

First line: Extrinsics - translation vector and rotation matrix
Second line: Intrinsics - focal length (normalized by dividing with the larger image dimension), distortion coefficients, pixel aspect ratio and principal point (in normalized coordinates [0,1])

If no cam files are available `makescene` behaves as before with a minimal overhead for the search of corresponding cam files (system calls are avoided by searching within the already parsed directory).

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -774,7 +774,8 @@ find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
     std::string prefix = remove_file_extension(dir[i].name);
 
     /* Look at files behind the given file until the prefix changes. */
-    for (std::size_t j = i + 1; j < dir.size(); ++j) {
+    for (std::size_t j = i + 1; j < dir.size(); ++j)
+    {
         if (dir[j].is_dir)
             continue;
 
@@ -788,14 +789,16 @@ find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
             break;
 
         std::string ext = fname.substr(pos + 1, fname.size());
-        if (util::string::lowercase(ext) == "cam") {
+        if (util::string::lowercase(ext) == "cam")
+        {
             *acfname = dir[j].get_absolute_name();
             return true;
         }
     }
 
     /* Look at files before the given file until the prefix changes. */
-    for (int j = i - 1; 0 <= j; --j) {
+    for (int j = i - 1; 0 <= j; --j)
+    {
         if (dir[j].is_dir)
             continue;
 
@@ -809,7 +812,8 @@ find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
             break;
 
         std::string ext = fname.substr(pos + 1, fname.size());
-        if (util::string::lowercase(ext) == "cam") {
+        if (util::string::lowercase(ext) == "cam")
+        {
             *acfname = dir[j].get_absolute_name();
             return true;
         }
@@ -934,7 +938,8 @@ import_images (AppSettings const& conf)
         view->set_name(remove_file_extension(fname));
 
         std::string acfname;
-        if (find_corresponding_cam_file(dir, i, &acfname)) {
+        if (find_corresponding_cam_file(dir, i, &acfname))
+        {
             try
             {
                 view->set_camera(load_cam_file(acfname));
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -766,14 +766,14 @@ find_max_scene_id (std::string const& view_path)
 
 /* ---------------------------------------------------------------- */
 
-/* Find corresponding cam file to file i in sorted directory. */
+/* Find corresponding cam file to image file i in sorted directory. */
 bool
 find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
     std::string * acfname)
 {
     std::string prefix = remove_file_extension(dir[i].name);
 
-    /* Look at files behind the given file until the prefix changes */
+    /* Look at files behind the given file until the prefix changes. */
     for (std::size_t j = i + 1; j < dir.size(); ++j) {
         if (dir[j].is_dir)
             continue;
@@ -794,7 +794,7 @@ find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
         }
     }
 
-    /* Look at files before the given file until the prefix changes */
+    /* Look at files before the given file until the prefix changes. */
     for (int j = i - 1; 0 <= j; --j) {
         if (dir[j].is_dir)
             continue;
@@ -913,6 +913,7 @@ import_images (AppSettings const& conf)
         std::string fname = dir[i].name;
         std::string afname = dir[i].get_absolute_name();
 
+        /* Ignore cam files. */
         if (fname.size() > 3 && util::string::right(fname, 4) == ".cam")
             continue;
 
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -766,6 +766,101 @@ find_max_scene_id (std::string const& view_path)
 
 /* ---------------------------------------------------------------- */
 
+/* Find corresponding cam file to file i in sorted directory. */
+bool
+find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
+    std::string * acfname)
+{
+    std::string prefix = remove_file_extension(dir[i].name);
+
+    /* Look at files behind the given file until the prefix changes */
+    for (std::size_t j = i + 1; j < dir.size(); ++j) {
+        if (dir[j].is_dir)
+            continue;
+
+        std::string const & fname = dir[j].name;
+
+        std::size_t pos = fname.find_last_of('.');
+        if (pos == std::string::npos)
+            continue;
+
+        if (fname.substr(0, pos) != prefix)
+            break;
+
+        std::string ext = fname.substr(pos + 1, fname.size());
+        if (util::string::lowercase(ext) == "cam") {
+            *acfname = dir[j].get_absolute_name();
+            return true;
+        }
+    }
+
+    /* Look at files before the given file until the prefix changes */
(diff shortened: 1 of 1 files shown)


apps/makescene/makescene.cc
```diff
@@ -766,6 +766,101 @@ find_max_scene_id (std::string const& view_path)
...
+            break;
+
+        std::string ext = fname.substr(pos + 1, fname.size());
+        if (util::string::lowercase(ext) == "cam") {
+            *acfname = dir[j].get_absolute_name();
+            return true;
+        }
+    }
+
+    /* Look at files before the given file until the prefix changes */
```
nmoehrle: A single loop would be more concise but I found it to be less readable.
```
for (int j = i + 1, step = 1; 0 <= j && j < (int)dir.size(); j += step)  {
    
    ...

    if (fname.substr(0, pos) != prefix) {                                                                                  
        if (step == 1)                                                                                                     
        {                                                                                                                  
            j = i;                                                                                                         
            step = -1;                                                                                                     
            continue;                                                                                                      
        }                                                                                                                  
        else                                                                                                               
        {                                                                                                                  
            return false;                                                                                                  
        }                                                                                                                  
    }

    ...
```

apps/makescene/makescene.cc
```diff
@@ -834,6 +932,26 @@ import_images (AppSettings const& conf)
         view->set_id(id);
         view->set_name(remove_file_extension(fname));
 
+        std::string acfname;
+        if (find_corresponding_cam_file(dir, i, &acfname)) {
```
simonfuhrmann: I don't understand your implementation of this function. Why not just

    string cam_filename = find_corresponding_cam_file(afname);
    if (!cam_filename.empty()) ...

apps/makescene/makescene.cc
```diff
@@ -766,6 +766,101 @@ find_max_scene_id (std::string const& view_path)
 
 /* ---------------------------------------------------------------- */
 
+/* Find corresponding cam file to file i in sorted directory. */
+bool
+find_corresponding_cam_file(util::fs::Directory const & dir, std::size_t i,
```
simonfuhrmann: Why is this iterating over other files and not just replacing the extension with `.cam` and checking if the file is there?

nmoehrle: To avoid system calls and check for `.cam` and `.CAM` files.

apps/makescene/makescene.cc
nmoehrle: To avoid system calls and string comparisons with the whole `dir`.

flanggut:
I'm not sure if i like this approach from a user's perspective. With this you automatically assume that .cam files in the directory are specifically there for MVE parsing. If I already have .cam files in my folder (for whatever reason) the import is triggered automatically and might just throw an error in this case without it being the user's fault. I would like to tigger this import myself, independent of the file structure I am working on, so at least I'd vote for a command line option. 
And if we're already having that, wouldn't it be easier to just add an option like `--import-cameras FILE` which reads a single file listing all filenames and the corresponding parameters?

lemony-fresh:
In relation to the argument that this import shouldn't happen implicitly: Don't the other bundle imports happen implicitly too? If you happen to have a synth_0.out in your folder (for whatever reason), makescene is using that.
BUT: This import should not be triggered by providing --images-only as its currently done, because this is more than images only. It's almost a full bundle, only the SfM feature point cloud is missing.

About putting all cameras into a single file: You would have to provide image file names in the cam file. The way it's handled now, you could e.g. remove images and corresponding cameras, or divide the scene into a training and a test set with simple file system operations. Putting all cams together in one file is problematic for the same reason why the data layout in MVE1 was inferior to the one in MVE2.

Last thought: This code should do undistorting because all the required data is there and it's consistent with what makescene does with other bundle formats. The only fundamental difference to other bundle formats is SfM points and their visibility and those are in no way related to undistortion, so it shouldn't be treated differently than other bundle formats.

flanggut:
If the user is providing custom distortion parameters he should actually do the undistortion himself.
So why even have all of this if you can just provide the information in the format of Noah's bundler output?

lemony-fresh:
As I wrote above:
"[With these .cam files] you could e.g. remove images and corresponding cameras, or divide the scene into a training and a test set with simple file system operations. Putting all cams together in one file is problematic for the same reason why the data layout in MVE1 was inferior to the one in MVE2."

I'm not saying there shouldn't be any bundle formats where all info is contained in a single file. There are people out there with synth files and they should be supported.
But having things distributed in multiple files has advantages as you all recognized when you switched to the MVE2 format. This file format is not contrived, Nils and I are currently using it and it has advantages.

nmoehrle:
Noah's bundler output does neither support pixel aspect ratios nor principal points. This excludes this format for all usages where the horizontal and vertical focal lengths differ.

flanggut:
> Putting  all cams together in one file is problematic for the same reason why the data layout in MVE1 was inferior to the one in MVE2.

No, this was mainly to reduce file system operations when changing/saving a single embedding.

>  you could e.g. remove images and corresponding cameras, or divide the scene into a training and a test set with simple file system operations

You could easily remove/add images from the single file configuration file and trigger different training/test sets by having multiple versions of this file. This would also avoid having to split up the actual dataset into multiple folders or creating TB of data by duplicating files.

nmoehrle:
> This would also avoid having to split up the actual dataset into multiple folders or creating TB of data by duplicating files.
How? `makescene` still creates a scene folder and copies all of the images.

flanggut:
> Noah's bundler output does neither support pixel aspect ratios nor principal points. This excludes this format for all usages where the horizontal and vertical focal lengths differ.

Yes. But I still don't like the idea of having yet another version of this.

We can already hard set these parameters by editing the `meta.ini` files. I feel like we should build on that. 

> How? `makescene` still creates a scene folder and copies all of the images.

I am talking about the original input data. `-i` imports all images inside a folder so you'd have to split up this original folder, right?

nmoehrle:
> I am talking about the original input data. -i imports all images inside a folder so you'd have to split up this original folder, right?

You could use symlinks for that but I get your point.

lemony-fresh:
Coming back to the MVE1 vs. MVE2 thing: The initial reason may have been a different one, but I liked the side effect that operations on the scene don't have to be done by touching individual file lines anymore.

nmoehrle:
Is there an bundle file format that supports pixel aspect ratios and principal points? Just looked at *.nvm* and *synth_0.out* and they do not.

*synth_<x>.out* cameras:
```
<cam 1 line 1> // Focal length, Radial distortion: f rd1 rd2
<cam 1 line 2> // Rotation matrix row 1: r11 r12 r13
<cam 1 line 3> // Rotation matrix row 2: r21 r22 r23
<cam 1 line 4> // Rotation matrix row 3: r31 r32 r33
<cam 1 line 5> // Translation vector: t1 t2 t3
```

*.nvm* cameras:
```
<Camera> = <File name> <focal length> <quaternion WXYZ> <camera center> <radial distortion> 0
```

So currently the only option, if you have different horizontal and vertical focal lengths, is to write out the scene structure yourself. I have done this multiple times with different python scripts but it takes some time.

simonfuhrmann:
Let's not talk about MVE1 vs. MVE2 file format change, but it's good that we can all agree that it's better now. I agree with @flanggut that we should not implicitly read `.cam` files without the users knowledge -- I suggested that and I realize it's a bad idea. But I can also see the convenience argument of having multiple `.cam` files.

Regarding distortion: MVE should not undistort the images. There are dozens of distortion models, and we don't want to support them. We should not even provide the distortion parameters in the file. Technically, distortion parameters are intrinsics, but for practically, we never use them except in SfM.

The only remaining question to me is, how can we import a scene with `.cam` files, and how should the file format be.

> In relation to the argument that this import shouldn't happen implicitly: Don't the other bundle imports happen implicitly too?

Actually, `makescene` is for explicitly importing bundles. The only exception is `--images-only`. In this spirit, maybe we should just provide a directory, and `makescene` should auto-detect that this is an "image-with-cam-files" scene, and simply do the right thing.

simonfuhrmann:
I added a makescene change that prepares for more general bundle format detection.

nmoehrle:
I like the idea of an automatic detection based on filename / directory content.

Regarding the file format I would vote for a format like the one originally proposed here without the distortion parameters. Although I see the benefits of a more or less self documenting format like ini files I would like to keep the format as simple as possible to allow the development of concise conversion scripts.

simonfuhrmann:
The format is up to you. But think about it. The format you proposed is basically not extensible, because the values are expected in a fixed order. In an ini-like format you can add and omit optional values later.

nmoehrle:
Appending values would be possible but see your concerns. I would like to hear some more opinions on this.

nmoehrle:
For this comment I will refer to the proposed old style format as .cam files and would like to compare them to the alternatives. The only alternative that I can currently see are .ini files - we already have a parser and I would like to dismiss json right away (completely ignoring xml):

- .ini files are extensible
- .ini files are basically self documenting
- .ini files have a larger overhead (file size and parsing)
- .cam files are minimal (for plain text files)
- parsing and writing .cam files is easier

If the files are just focused on camera parameters are there any that could be added?

simonfuhrmann:
For texturing, for example, you could add application specific information (that will be ignored by MVE). 

Also, I don't think that file size and parsing overhead is a good argument. Reading and parsing the image is on a completely different order of magnitude. And ".cam files are minimal" is just a statement, not sure what you mean by that. And I would actually argue against that `.ini` files are harder to write. In my experience, writing

    focal_length = 0.98
    translation = 0.1 0.3 0.5

is actually "easier" than writing a bunch of numbers without self-documenting meaning in the correct order.

simonfuhrmann:
Closing due to inactivity.

PR #292, opened 2017-01-11
APPS: Fix makescene import of 16 bit tiff files

nmoehrle:
Importing 16 bit images fails due to exceptions thrown when trying to load them as 8 bit images. This fix ignores these exceptions within `load_8bit_image` following the pattern used in `load_16bit_image` and `load_float_image`.

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -123,20 +123,13 @@ load_8bit_image (std::string const& fname, std::string* exif)
     std::string lcfname(util::string::lowercase(fname));
     std::string ext4 = util::string::right(lcfname, 4);
     std::string ext5 = util::string::right(lcfname, 5);
-    try
-    {
-        if (ext4 == ".jpg" || ext5 == ".jpeg")
-            return mve::image::load_jpg_file(fname, exif);
-        else if (ext4 == ".png" ||  ext4 == ".ppm"
-            || ext4 == ".tif" || ext5 == ".tiff")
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
Line 620 makes use of the exception.

nmoehrle:
True - missed that...

Should I `try {  } catch (...) throw;` instead of removing the try catches entirely?

simonfuhrmann:
No, just return `nullptr` as before and check for `nullptr` in line 620.

nmoehrle:
Thought about that but we then cannot give any information on what went wrong in 620...

No such file, permission denied, etc.

Your call

simonfuhrmann:
Sure. But since you're not GIVING any information, I'd go for the version that is cleaner. Which is the `nullptr` check.

nmoehrle:
Aren't `FileException`s that are raise in [`mve::image::load_file`](https://github.com/nmoehrle/mve/blob/master/libs/mve/image_io.cc#L57) thrown as `Exception`s in
[image_io.cc#L100](https://github.com/nmoehrle/mve/blob/master/libs/mve/image_io.cc#L100) and then ultimately printed in [makescene.cc#L644](https://github.com/nmoehrle/mve/blob/master/apps/makescene/makescene.cc#L644)?

simonfuhrmann:
Yes. But non of that catches you introduced print them.

nmoehrle:
None of the catches I introduced but the original one that I missed when I first opened this PR.

PR #299, opened 2017-02-10
Add recipe to build a snap package

mardy:
Snap is a cross-distribution packaging format for Linux (see https://snapcraft.io).

By running the `snapcraft` command on this repository, it will be possible to generate a `.snap` package of mve. Generating the snap file has been tested on Ubuntu 16.04, yet the resulting package should work on any snap-enabled distribution.

I'm taking the liberty of uploading the mve snap into the snap store, so that people will be able to install it just by typing
```
    snap install mve
```
If you let me know your e-mail address, I will add you as a maintainer of the package, so you'll be able to upload new versions in the snap store, in case I'm unavailable to do it.


simonfuhrmann:
Hi. We are still debating whether this should be into the repo or not. But two points:

 * I think it should go into a sub-directory, like `dist/`, in case we want to add more packaging related files
 * A readme file should be added to document what this is and how it is used

Thoughts?

mardy:
Hi Simon! Well, these files already are into their `snap` subdirectory, but of course they can be moved deeper in the tree, that's not going to be a problem. I will move them into `dist`, then.
I will also add a readme.

Please let me know your other concerns. Meanwhile, let me explain why I contributed this: I've been testing a few different SfM programs to prepare a quick video where I do a comparison of them for the casual user (I can already tell you, that mve scores pretty high :-) ). Now, many of these programs are not easy to install on Linux so, since I had to compile them anyway, I thought of building them as snap packages, in order to reduce the trouble for others users who want to try them out but are not so comfortable with compilation and stuff.
Unfortunately I don't have time to follow the development of these projects (and my e-mail would just explode if I subscribed to all of them), so I won't be actively watching this. I will be very happy to help in case you need assistance to build snap packages in the future, but I won't be making snaps for the new releases spontaneously, unless some of you remind me of that.

I'm telling you this because I think it can affect your decision; basically, my hope is that you'll like the idea and start distributing mve as a snap yourself.

If I may add one point in favor of snaps, is that given that they bundle all their dependency, when someone reports a bug and tells you the version number he has installed, you'll be able to tell exactly how the package was built, with what libraries and stuff, so replicating the issue it will be easier.

simonfuhrmann:
Sure that sounds reasonable. Since MVE doesn't change rapidly, updating the snap once in a while is fine. Where are these packages? Is there a website that lists them?

For now, I think we should move `/snap/` to `/dist/snap/` and add a `README.txt` or `README.md` to explain what this is doing, how to create new snaps and where they are located. A few links are fine.

mardy:
Hi! I've moved the files and added a README.

AFAIK, there isn't an official snap website at the moment, but the server exposes a REST API, so anyone can build a website interfacing to it. Here's the one I know of:

  https://uappexplorer.com/app/mve.mardy

I realize just now that the app is appearing under my name, which is not what I'd like. I suggest that you register an account for the GRIS (maybe using the mve@GRIS e-mail address you have) and then add yourself and other members of the team (and me, if you'd like me to publish the packages) as collaborators.
This can be done by following the instructions here:
https://snapcraft.io/docs/build-snaps/publish

PR #302, opened 2017-02-23
Fix matrix computation of rephotographer (OpenGL convention)

nmoehrle:
The current implementation uses the computer vision camera convention (looking into the direction of z) and compensates with flips of y and z in the projection. This results in incorrect normals during shading.


nmoehrle:
I would like to see the functionality to create or fill a ogl::Camera in the libraries. If you agree should they go into [math/matrix_tools](https://github.com/simonfuhrmann/mve/blob/master/libs/math/matrix_tools.h) like `matrix_gl_projection` or rather into the `ogl` or `mve` lib?

simonfuhrmann:
You can move the functionality to `math/matrix_tools.h`.

PR #304, opened 2017-02-24
Extract gl matrix creation

nmoehrle:
I tried to put these functions into `math/matrix_tools.h` but the creation of the projection matrix requires a lot of information from `mve::CameraInfo` and relies on our camera convention so it felt more natural to put it into `mve::CameraInfo` directly.

### DIFF
--- a/libs/mve/camera.cc
+++ b/libs/mve/camera.cc
@@ -69,6 +69,17 @@ CameraInfo::fill_world_to_cam (float* mat) const
 /* ---------------------------------------------------------------- */
 
 void
+CameraInfo::fill_gl_viewtrans (float* mat) const
+{
+    mat[0]  =  rot[0]; mat[1]  =  rot[1]; mat[2]  =  rot[2]; mat[3]  =  trans[0];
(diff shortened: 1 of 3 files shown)


libs/mve/camera.cc
```diff
@@ -135,6 +146,37 @@ CameraInfo::fill_calibration (float* mat, float width, float height) const
 /* ---------------------------------------------------------------- */
 
 void
+CameraInfo::fill_gl_projection (float * mat, float width, float height,
```
simonfuhrmann: Remove space before pointer

libs/mve/camera.h
```diff
@@ -134,6 +134,17 @@ struct CameraInfo
...
 
     /**
+     * Stores OpenGl view transformation 4x4 matrix in array pointed to by mat.
+     */
+    void fill_gl_viewtrans (float* mat) const;
+
+    /**
+     * Stores OpenGl projection 4x4 matrix in array pointed to by mat.
+     */
+    void fill_gl_projection (float * mat, float width, float height,
```
simonfuhrmann: remove space before pointer

PR #306, opened 2017-02-27
Fix quaternion to rotation matrix conversion (wrong matrix convention)

nmoehrle:
The rotation matrix is written out as column major, resulting in transposed rotations.


simonfuhrmann:
Nice. I hope this works.

PR #310, opened 2017-03-09
SFM: Improve worst case runtime of track propagation v2

nmoehrle:
I encountered a dataset where the track propagation did not complete after 24 hours. Unifying in the larger of both tracks and effectively reserving the required memory beforehand with insert reduced the runtime to under 15 seconds.

### DIFF
--- a/libs/sfm/bundler_tracks.cc
+++ b/libs/sfm/bundler_tracks.cc
@@ -17,6 +17,35 @@
 SFM_NAMESPACE_BEGIN
 SFM_BUNDLER_NAMESPACE_BEGIN
 
+/*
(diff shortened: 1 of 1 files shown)


nmoehrle:
Coffee break discussions...

simonfuhrmann:
I still can't believe that this change reduces time so much. Does this somehow fix an infinite loop bug?

nmoehrle:
I don't think so I was monitoring the process and it did make progress but in a painstakingly slow pace and I don't know which loop would be a candidate. I am happy to provide the prebundle but it has more than 11Gb (> 1400 views).

simonfuhrmann:
Don't worry. Good that his fixes it.

PR #312, opened 2017-03-10
Add survey point support

nmoehrle:
I am happy to rename any of the newly introduced concepts.

Alternatives for survey point: tie point, anchor, ground control point, control point
Alternatives for survey point observation: marker, control feature, annotation

@seepa thanks for your help :-)

### DIFF
--- a/libs/math/matrix_tools.h
+++ b/libs/math/matrix_tools.h
@@ -115,6 +115,13 @@ Vector<T,N>
 matrix_get_diagonal (Matrix<T,N,N> const& mat);
 
 /**
+ * Calculates the trace of the given matrix.
+ */
+template <typename T, int N>
+T
+matrix_trace(math::Matrix<T, N, N> const& mat);
+
+/**
  * Calculates the determinant of the given matrix.
  * This is specialized for 1x1, 2x2, 3x3 and 4x4 matrices only.
  */
@@ -356,6 +363,16 @@ matrix_get_diagonal (Matrix<T,N,N> const& mat)
     for (int i = 0, j = 0; i < N*N; i += N+1, j += 1)
         diag[j] = mat[i];
     return diag;
+}
+
+template <typename T, int N>
+inline T
+matrix_trace(math::Matrix<T, N, N> const& mat)
+{
+    T ret = T(0.0);
+    for (int i = 0; i < N; ++i)
+        ret += mat(i, i);
+    return ret;
 }
 
 template <typename T>
--- a/libs/math/matrix_tools.h
+++ b/libs/math/matrix_tools.h
@@ -369,9 +369,9 @@ template <typename T, int N>
 inline T
 matrix_trace(math::Matrix<T, N, N> const& mat)
 {
-    T ret = T(0.0);
-    for (int i = 0; i < N; ++i)
-        ret += mat(i, i);
+    T ret(0.0);
+    for (int i = 0; i < N * N; i += N + 1)
+        ret += mat[i];
     return ret;
 }
 
--- a/libs/math/transform.h
+++ b/libs/math/transform.h
@@ -15,16 +15,32 @@
 
 MATH_NAMESPACE_BEGIN
 
+/*
+ * Determines the transformation between two lists of corresponding points.
+ * Minimizing sum_i (R * s * p0[i] + t - p1[i])^2
+ */
 template <typename T, int N>
 bool
-determine_transform(std::vector<math::Vector<T, N> > const& p0,
-    std::vector<math::Vector<T, N> > const& p1,
+determine_transform(std::vector<math::Vector<T, N>> const& p0,
+    std::vector<math::Vector<T, N>> const& p1,
+    math::Matrix<T, N, N>* rot, T* scale, math::Vector<T, N>* trans);
+
+MATH_NAMESPACE_END
+
+/* ------------------------ Implementation ------------------------ */
+
+MATH_NAMESPACE_BEGIN
+
+template <typename T, int N>
+bool
+determine_transform(std::vector<math::Vector<T, N>> const& p0,
+    std::vector<math::Vector<T, N>> const& p1,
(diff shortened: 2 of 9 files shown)


flanggut:
A quick note: I suggested the way this is integrated into the BA - it's not the most elegant way to do this but it avoids major changes in the BA data structures and the handling of the jacobian/hessian/precondition matrices.

libs/math/matrix_tools.h
```diff
@@ -358,6 +365,16 @@ matrix_get_diagonal (Matrix<T,N,N> const& mat)
     return diag;
 }
 
+template <typename T, int N>
+inline T
+matrix_trace(math::Matrix<T, N, N> const& mat)
+{
+    T ret = T(0.0);
```
simonfuhrmann: T ret(0.0);

libs/math/matrix_tools.h
```diff
@@ -358,6 +365,16 @@ matrix_get_diagonal (Matrix<T,N,N> const& mat)
...
+    for (int i = 0; i < N; ++i)
+        ret += mat(i, i);
```
simonfuhrmann: Instead of using operator(), I think this can be made more efficient using operator[] and adding N+1 in the for increment.

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+
+#ifndef MATH_TRANSFORM_HEADER
+#define MATH_TRANSFORM_HEADER
+
+#include "math/matrix.h"
+#include "math/matrix_svd.h"
+
+MATH_NAMESPACE_BEGIN
+
+template <typename T, int N>
```
simonfuhrmann: Can you please add a function declaration with a comment what this function is doing?
An then separate the declaration from the definition with `/* --- Implementation --- */` like in the other header files.

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+bool
+determine_transform(std::vector<math::Vector<T, N> > const& p0,
```
simonfuhrmann: With C++11, there is no need for the extra space before `>`. Please fix in this file.
Later, we might want to go over MVE and fix it everywhere.

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+    std::vector<math::Vector<T, N> > const& p1,
+    math::Matrix<T, N, N>* rot, T* scale, math::Vector<T, N>* trans)
+{
+    if (p0.size() != p1.size())
+        throw std::invalid_argument("Dimension size mismatch");
+
+    std::size_t num_correspondences = p0.size() & p1.size();
```
simonfuhrmann: Wow, what is this?

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+        throw std::invalid_argument("At least three correspondences required");
+
+    /* Calculate centroids. */
+    math::Vector<T, N> c0(0.0), c1(0.0);
+    for (std::size_t i = 0; i < num_correspondences; ++i)
+    {
+        c0 += p0[i];
+        c1 += p1[i];
+    }
+    c0 /= num_correspondences;
```
simonfuhrmann: `static_cast<T>(num_correspondences)`

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+    c1 /= num_correspondences;
+
+    /* Calculate covariance and variance. */
+    T sigma2 = 0.0;
```
simonfuhrmann: `T sigma2(0.0)`
Depending on `T`, assignment might not work.

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+    math::Matrix<T, N, N> cov(0.0);
+    for (std::size_t i = 0; i < num_correspondences; ++i)
+    {
+        math::Vector<T, N> pc0 = p0[i] - c0;
+        math::Vector<T, N> pc1 = p1[i] - c1;
+        cov += math::Matrix<T, N, 1>(pc0.begin()) *
+            math::Matrix<T, 1, N>(pc1.begin());
+        sigma2 += pc0.square_norm();
+    }
+    cov /= num_correspondences;
```
simonfuhrmann: cast

libs/math/transform.h
```diff
@@ -0,0 +1,90 @@
...
+    sigma2 /= num_correspondences;
+
+    /* Determine rotation and scale. */
+    math::Matrix<T, N, N> U, S, V;
+    math::matrix_svd(cov, &U, &S, &V);
+    math::Matrix<T, N, N> R = V * U.transposed();
+
+    if (S(N - 1, N -1) < std::numeric_limits<T>::epsilon())
```
simonfuhrmann: `N - 1`
Also, are you sure `std::numeric_limits<T>::epsilon()` is actually what you want? This number is really small and usually not what you want for epsilon checks. `math/defines.h` has some more choices.

libs/math/transform.h
nmoehrle: Isn't the cast implicit?

libs/math/transform.h
nmoehrle: Sorry habit both values are the same -> their binary representations is the same so the statement is will give the same result as ` std::size_t num_correspondences = p0.size();` but show the intent that both should be the same.

libs/math/transform.h
nmoehrle: It's essentially just checking if the co-variance matrix was rank deficient, in this case the entry should be zero. @dthuerck any suggestions for a better threshold?

dthuerck:
Better take the SVD-internal Zero threshold (whatever that was...). That way, the deflation algorithm will be correct for your epsilon....

On Mar 10, 2017 8:17 PM, Simon Fuhrmann <notifications@github.com> wrote:

Merged #312<https://github.com/simonfuhrmann/mve/pull/312>.

PR #313, opened 2017-03-12
Parallel imba

nmoehrle:
With these patches the runtime of a 113 image dataset (with given prebundle) could be reduced from thirteen to eight minutes. The computer used had a Intel(R) Core(TM) i7 CPU 960 @ 3.20GHz (4 cores, 8 threads). The single threaded matrix multiplication was the biggest bottleneck of bundling I did not evaluate the other parts separately.

### DIFF
--- a/libs/sfm/ba_sparse_matrix.h
+++ b/libs/sfm/ba_sparse_matrix.h
@@ -10,6 +10,7 @@
 #ifndef SFM_SPARSE_MATRIX_HEADER
 #define SFM_SPARSE_MATRIX_HEADER
 
+#include <thread>
 #include <stdexcept>
 #include <vector>
 #include <algorithm>
@@ -54,6 +55,8 @@ public:
     SparseMatrix transpose (void) const;
     SparseMatrix subtract (SparseMatrix const& rhs) const;
     SparseMatrix multiply (SparseMatrix const& rhs) const;
+    SparseMatrix sequential_multiply (SparseMatrix const& rhs) const;
+    SparseMatrix parallel_multiply (SparseMatrix const& rhs) const;
     DenseVector<T> multiply (DenseVector<T> const& rhs) const;
     SparseMatrix diagonal_matrix (void) const;
 
@@ -121,6 +124,7 @@ template <typename T>
 void
 SparseMatrix<T>::reserve (std::size_t num_elements)
 {
+    this->inner.reserve(num_elements);
     this->values.reserve(num_elements);
 }
 
@@ -256,6 +260,17 @@ template <typename T>
 SparseMatrix<T>
 SparseMatrix<T>::multiply (SparseMatrix const& rhs) const
 {
+#ifdef _OPENMP
+    return this->parallel_multiply(rhs);
+#else
+    return this->sequential_multiply(rhs);
+#endif
+}
+
+template <typename T>
+SparseMatrix<T>
+SparseMatrix<T>::sequential_multiply (SparseMatrix const& rhs) const
+{
     if (this->cols != rhs.rows)
         throw std::invalid_argument("Incompatible matrix dimensions");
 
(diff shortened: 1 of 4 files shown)


libs/sfm/ba_sparse_matrix.h
```diff
@@ -256,6 +259,17 @@ template <typename T>
 SparseMatrix<T>
 SparseMatrix<T>::multiply (SparseMatrix const& rhs) const
 {
+#ifdef _OPENMP
+    return this->parallel_multiply(rhs);
```
nmoehrle: Parallelization only makes sense for reasonable large matrices, we could introduce a switch here.

libs/sfm/ba_sparse_matrix.h
```diff
@@ -54,6 +55,8 @@ class SparseMatrix
     SparseMatrix transpose (void) const;
     SparseMatrix subtract (SparseMatrix const& rhs) const;
     SparseMatrix multiply (SparseMatrix const& rhs) const;
+    SparseMatrix sequential_multiply (SparseMatrix const& rhs) const;
```
simonfuhrmann: Should this be private/protected?

libs/sfm/ba_sparse_matrix.h
simonfuhrmann: Is performance degrading a lot when matrices are small?

libs/sfm/ba_sparse_matrix.h
```diff
@@ -299,6 +313,88 @@ SparseMatrix<T>::multiply (SparseMatrix const& rhs) const
...
+    if (this->cols != rhs.rows)
+        throw std::invalid_argument("Incompatible matrix dimensions");
+
+    std::size_t nnz = this->num_non_zero() + rhs.num_non_zero();
+    SparseMatrix ret(this->rows, rhs.cols);
+    ret.reserve(nnz);
+    std::fill(ret.outer.begin(), ret.outer.end(), 0);
+
+    std::size_t num_threads = std::max(1u,
+        std::thread::hardware_concurrency());
```
simonfuhrmann: Isn't OpenMP using this many threads by default?

libs/sfm/ba_sparse_matrix.h
```diff
@@ -299,6 +313,88 @@ SparseMatrix<T>::multiply (SparseMatrix const& rhs) const
...
+        thread_inner.reserve(nnz / num_threads);
+
+#pragma omp for ordered schedule(static, 1)
+        for (std::size_t chunk = 0; chunk < num_threads; ++chunk)
+        {
+            /* Necessary if compiled without openmp. */
+            thread_inner.clear();
+            thread_values.clear();
+
+            std::size_t begin = chunk * chunk_size;
```
simonfuhrmann: You can make these (and some variables) const to prevent accidental changing.

libs/sfm/ba_sparse_matrix.h
nmoehrle: Changed.

libs/sfm/ba_sparse_matrix.h
nmoehrle: Done.

libs/sfm/ba_sparse_matrix.h
nmoehrle: Didn't have time to test it yet.

libs/sfm/ba_linear_solver.cc
```diff
@@ -72,7 +72,12 @@ namespace
             math::Matrix<double, 3, 3> rot;
             for (int i = 0; i < 9; ++i)
                 rot[i] = *(iter++);
-            rot = math::matrix_inverse(rot);
+
+            double det = math::matrix_determinant(rot);
+            if (std::abs(det) < MATH_DBL_EPS)
```
simonfuhrmann: There is a math define which is more efficient that `std::abs`, please use `MATH_EPSILON_EQ` instead.

simonfuhrmann:
Ok, most of this stuff is nice, but please don't pile more commits on this PR. Let's do this one after the other, and let's first start with JUST the code that has most impact, which I guess is parallelization of the sparse matrix multiplication. Create one PR with that, and I will pull once everybody is happy. Then we move on the the next piece of code.

nmoehrle:
Should I split this into three different PRs?

simonfuhrmann:
Just start with the sparse matrix multiplication first. Then I am interested if some improvement numbers for the the parallel Jacobian. I might be wrong, it seems Jacobian computation doesn't take very long?

nmoehrle:
Some preliminary results for the runtime improvements on a Intel(R) Xeon(R) CPU E5-2687W v3 @ 3.10GHz (40 cores). Runtime on the Citywall dataset with 363 views and a precomputed prebundle (673619 feature tracks).

c21ee765: serial ~ 418 min
81e697e6: + parallel sparse matrix multiplication ~ 356 min (~ 15% runtime reduction)
02b9754a: + parallel jacobian ~ 337 min (~ 5% runtime reduction)
81e697e6: + parallel ransac 3D pose ~ 226 min (~ 33% runtime reduction)

Total runtime reduction ~ 46%

```
grep "SfM reconstruction took " *.log
02b9754a.log:SfM reconstruction took 20234374 ms.
5cc72384.log:SfM reconstruction took 13582638 ms.
81e697e6.log:SfM reconstruction took 21365347 ms.
c21ee765.log:SfM reconstruction took 25102063 ms.
```

simonfuhrmann:
Closing due to having separate PRs.

PR #316, opened 2017-03-20
SFM: Fix bug with zero hessian (constant points)

nmoehrle:


libs/sfm/ba_linear_solver.cc
```diff
@@ -46,7 +46,10 @@ namespace
 
             iter = iter_backup;
             for (int i = 0; i < bs2; ++i)
-                *(iter++) = matrix_block[i];
+                if (std::isfinite(matrix_block[i]))
+                    *(iter++) = matrix_block[i];
+                else
+                    *(iter++) = 0;
```
simonfuhrmann: 0.0

PR #319, opened 2017-03-28
Snap: add removable-media interface

mardy:
This allows the user to connect the "removable-media" interface and
grant mve access to the files located in their removable media.


simonfuhrmann:
Why do we require `unity7` for UMVE?

PR #325, opened 2017-04-11
SFM: Paralellize sparse matrix multiplication

nmoehrle:
Split from #313

### DIFF
--- a/libs/sfm/ba_sparse_matrix.h
+++ b/libs/sfm/ba_sparse_matrix.h
@@ -10,6 +10,7 @@
 #ifndef SFM_SPARSE_MATRIX_HEADER
 #define SFM_SPARSE_MATRIX_HEADER
 
+#include <thread>
 #include <stdexcept>
 #include <vector>
 #include <algorithm>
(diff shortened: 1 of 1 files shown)


libs/sfm/ba_sparse_matrix.h
```diff
@@ -299,6 +314,88 @@ SparseMatrix<T>::multiply (SparseMatrix const& rhs) const
...
+    std::size_t nnz = this->num_non_zero() + rhs.num_non_zero();
+    SparseMatrix ret(this->rows, rhs.cols);
+    ret.reserve(nnz);
+    std::fill(ret.outer.begin(), ret.outer.end(), 0);
+
+    std::size_t const chunk_size = 64;
+    std::size_t const num_chunks = ret.cols / chunk_size
+         + (ret.cols % chunk_size != 0);
+    std::size_t const max_threads = std::max(1u,
+        std::thread::hardware_concurrency());
```
nmoehrle: It appears that using all available threads does not make sense on CPUs with hyper threading but I could not find a way to determine the number of physical cores.

PR #326, opened 2017-04-11
SFM: Parallelize jacobian computation

nmoehrle:
Split from #313

### DIFF
--- a/libs/sfm/bundle_adjustment.cc
+++ b/libs/sfm/bundle_adjustment.cc
@@ -7,6 +7,7 @@
  * of the BSD 3-Clause license. See the LICENSE.txt file for details.
  */
 
+#include <thread>
 #include <iostream>
 #include <iomanip>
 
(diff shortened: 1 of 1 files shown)


flanggut:
I checked the details: The move constructor should be generated by default (see [reference](http://en.cppreference.com/w/cpp/language/move_constructor) ) and if it exists it is also called instead of the copy constructor. So this should be fine as it is.

simonfuhrmann:
It is generated by default under some circumstances, but it is never called for a regular assignment. You need `std::move`. But as already mentioned, since the class contains only primitives, there is really no advantage in moving.

flanggut:
Try this https://gist.github.com/flanggut/f30921d0422ab65d69b955556dc3a81f
For me it says MOVE.

simonfuhrmann:
Ok, you're right. Because it's a temporary.

flanggut:
Should we merge this?

simonfuhrmann:
Your call. If it improves performance and is worth having, merge it! ;-)

PR #328, opened 2017-04-19
Normals

seepa:
This PR implements face and vertex normal rendering in UMVE. I have refactored shader loading to support vert/geom/frag shaders with different basenames and implemented two geom shaders for vertex/face normal generation as well as one simple vert and frag shader. The length of the normals is calculated with a 2^x function and the default size looks good for meshes in the unit cube.

### DIFF
--- a/libs/ogl/shader_program.cc
+++ b/libs/ogl/shader_program.cc
@@ -101,18 +101,24 @@ ShaderProgram::compile_shader (GLuint shader_id, std::string const& code)
 bool
 ShaderProgram::try_load_all (std::string const& basename)
 {
-    std::string vert_filename = basename + ".vert";
-    std::string geom_filename = basename + ".geom";
-    std::string frag_filename = basename + ".frag";
+    return try_load_specific(basename + ".vert", basename + ".geom",
+        basename + ".frag");
+}
 
+bool
+ShaderProgram::try_load_specific (std::string const& vert_filename,
+    std::string const& geom_filename, std::string const& frag_filename)
+{
     if (!util::fs::file_exists(vert_filename.c_str())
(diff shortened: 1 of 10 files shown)


libs/ogl/shader_program.cc
```diff
@@ -101,18 +101,27 @@ ShaderProgram::compile_shader (GLuint shader_id, std::string const& code)
...
-    std::string geom_filename = basename + ".geom";
-    std::string frag_filename = basename + ".frag";
+    return try_load_specific(basename, basename, basename);
+}
+
+bool
+ShaderProgram::try_load_specific (std::string const& vert_basename,
+    std::string const& geom_basename, std::string const& frag_basename)
+{
+    std::string vert_filename = vert_basename + ".vert";
```
simonfuhrmann: For this method, we should not append the extension. I.e., construct the filename from the basename in `try_load_all`, and rename the arguments to `vert_filename`, etc, here.

vikasTmz:
@seepa I'm getting `undefined reference` ERROR at  `this->normals_face_shader->try_load_specific()`.

seepa:
yes, I'm getting the same. I'll try to find out why this suddenly doesn't work anymore.

seepa:
@vikasTmz could you try again, please?

I was able to track down the bug with glGetProgramInfoLog(). There was an error when linking vertex and geometry shader: the vertex shader had a vec4 color as output while the geom shader expected a vec3 color as input. For some reason, my PC at work did not complain about this mistake (OpenGL 3.3). I have now removed the color completely, as it is set in the fragment shader manually.

simonfuhrmann:
Closing old PRs.

PR #332, opened 2017-04-26
BA: Add support for fixed intrinsics

flanggut:


libs/sfm/bundle_adjustment.cc
```diff
@@ -629,17 +662,28 @@ void
...
+    int const offset = this->opts.fixed_intrinsics ? 3 : 0;
+    out->translation[0] = cam.translation[0] + update[3 - offset];
+    out->translation[1] = cam.translation[1] + update[4 - offset];
+    out->translation[2] = cam.translation[2] + update[5 - offset];
 
     double rot_orig[9];
     std::copy(cam.rotation, cam.rotation + 9, rot_orig);
     double rot_update[9];
-    this->rodrigues_to_matrix(update + 6, rot_update);
+    this->rodrigues_to_matrix(update + 6 - offset, rot_update);
```
simonfuhrmann: How about setting `offset` to either 6 or 3 based on `this->opts.fixed_intrinsics` and just do `update + offset` here?

flanggut: Yes that's better, I'll change it

PR #333, opened 2017-04-26
SFM: Find outliers during track triangulation

flanggut:
If a track can be triangulated with a subset of cameras
the other cameras, causing large errors, are now separated
from the track and a new track is generated. This helps if
false matches connect two otherwise fine tracks.

### DIFF
--- a/libs/sfm/triangulate.cc
+++ b/libs/sfm/triangulate.cc
@@ -87,74 +87,93 @@ is_consistent_pose (Correspondence2D2D const& match,
 bool
 Triangulate::triangulate (std::vector<CameraPose const*> const& poses,
     std::vector<math::Vec2f> const& positions,
-    math::Vec3d* track_pos,
-    Statistics* stats) const
+    math::Vec3d* track_pos, Statistics* stats,
+    std::vector<std::size_t>* outliers) const
 {
     if (poses.size() < 2)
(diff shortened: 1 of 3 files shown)


libs/sfm/triangulate.cc
```diff
@@ -87,74 +87,93 @@ is_consistent_pose (Correspondence2D2D const& match,
...
+        if (stats != nullptr)
+            stats->num_too_small_angle += 1;
+        return false;
     }
-    average_error /= static_cast<double>(poses.size());
 
-    /* Reject track if the reprojection error is too large. */
-    if (average_error > this->opts.error_threshold)
+    /* Check if required number of inliers is found */
+    if (poses.size() < best_outliers.size() + this->opts.min_num_views)
```
nmoehrle: Wouldn't 'poses.size() - best_outliers.size() < this->opts.min_num_size' be more legible or am I missing a unsigned case when 'best_outliers.size()' is bigger than 'poses.size()' which should not be possible anyway.

flanggut: I guess it doesn't make a difference here as `outliers.size()` should always be smaller than `poses.size()`. I am just trying to avoid subtraction with unsigned types in general.

simonfuhrmann: I would avoid subtraction on unsigned types.

PR #337, opened 2017-05-11
Clean up of Scene and View code

LinusVanElswijk:
I'm trying making the code easier to read and maintain.
This PR reflects my initial work in that direction and changes only the Scene and View files:

- Replacing low level constructs by higher level constructs made available in C++11 where appropriate.
- Small changes to functions or new functions to reduce the number of repeating patterns in code.
- Commented out code has been removed. Most of these where  console output.

The only functional change is that the Scene(void) constructor has been made private, since
the documentation seems to suggest that it should not be used and it is not used anywhere outside of the class definition.

None of the touched code it hot code so there should not be any concerns about performance here.

### DIFF
--- a/libs/mve/scene.cc
+++ b/libs/mve/scene.cc
@@ -30,18 +30,9 @@ Scene::load_scene (std::string const& base_path)
 /* ---------------------------------------------------------------- */
 
 void
-Scene::save_scene (void)
-{
-    this->save_bundle();
-    this->save_views();
-}
-
-/* ---------------------------------------------------------------- */
-
-void
 Scene::save_bundle (void)
 {
-    if (this->bundle == nullptr || !this->bundle_dirty)
+    if (!(this->bundle && this->bundle_dirty))
         return;
     std::string filename = util::fs::join_path(this->basedir, "synth_0.out");
     save_mve_bundle(this->bundle, filename);
@@ -54,25 +45,11 @@ void
 Scene::save_views (void)
 {
     std::cout << "Saving views to MVE files..." << std::flush;
-    for (std::size_t i = 0; i < this->views.size(); ++i)
-        if (this->views[i] != nullptr && this->views[i]->is_dirty())
-            this->views[i]->save_view();
+
+    for (auto& view : views)
+        if (view && view->is_dirty())
+            view->save_view();
     std::cout << " done." << std::endl;
-}
-
-/* ---------------------------------------------------------------- */
-
-bool
-Scene::is_dirty (void) const
-{
-    if (this->bundle_dirty)
-        return true;
-
-    for (std::size_t i = 0; i < this->views.size(); ++i)
-        if (this->views[i] != nullptr && this->views[i]->is_dirty())
-            return true;
-
-    return false;
 }
 
 /* ---------------------------------------------------------------- */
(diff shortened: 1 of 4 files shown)


simonfuhrmann:
Hi Linus,

thanks for your contribution. However, I will not merge this. I think some of the changes are useful and I would merge them (making the constructor private for example). However, I think most "style" related changes don't make the code more readable, but IMO have the opposite effect.

A few examples: 

* Relying on boolean operators on pointers to test for `nullptr` is something that removes valuable type information when reading the code. Many things evaluate to a bool, e.g., a bool, an int, a pointer, `std::function`, etc. Being explicit documents the type.
* The same goes for `auto`. When used carefully, `auto` can be a useful C++ feature (e.g., for iterators), but in most other situations using auto is dangerous it makes the code completely unreadable.
* The use of `++var` as a statement is debatable, but I think `var += 1` just looks more like a proper statement.

LinusVanElswijk:
Thank you for the feedback. Do you see any benefit in merging this PR after the style changes you mentioned are reverted? This would restrict the changes to mostly
- replacement of explicit index loops with for each style loops ([example](https://github.com/simonfuhrmann/mve/pull/337/files#diff-85f2de38ac41530b485e1c2bf7d459c5R45))
- removal of commented out code
- reduction of repeated code patterns ([example](https://github.com/simonfuhrmann/mve/pull/337/files#diff-31cc4db4545d0f9f07ec3a4ee6779bceR441))
- more concise formulations ([example1](https://github.com/simonfuhrmann/mve/pull/337/files#diff-31cc4db4545d0f9f07ec3a4ee6779bceR288), [example2](https://github.com/simonfuhrmann/mve/pull/337/files#diff-85f2de38ac41530b485e1c2bf7d459c5R86)) 

Since there is no style guide available, determining the preferred way of writing often comes down to guessing. I could open an issue for that and use and devote a bit of time on a style guide if that would be appreciated.

I disagree that committing to explicit types without good reason has any beneficial impact over programming against their interfaces, but I am willing to adapt to whatever is preferred. :upside_down_face: 
<details>
<summary>However, if you are interested in my reasoning...</summary>
When explicit types are repeated, the burden of keeping all these types in sync with each-other is put on the programmer, instead of the compiler. In many cases, error in repeating explicit types will result in unwanted implicit conversions. This happens in the original code, where the return type cache_cleanup is incorrectly assumed to be std::size_t.

Both `if (value) ...` and `if (value == 0.0) ...` would be perfectly readable if put in the correct context and with a more descriptive variable name. The correct context of the first is when value is a boolean or if there may not be a value. The correct context to use the second is when `value` certainly holds a value that can be compared against a real number. I do not care if the type of value is actually double or DegreesCelsius.

This follows the Almost Always Auto guideliness ([link](http://cginternals.github.io/guidelines/articles/almost-always-auto/ ) and [link](https://herbsutter.com/2013/08/12/gotw-94-solution-aaa-style-almost-always-auto/)).
</details>

simonfuhrmann:
I am fine with the changes you mentioned:

* The code has been started in a non-C++11 environment, and converting some constructs to the new style is appreciated, especially more towards using the range-based for loop (if used without `auto`).
* Commented code can most certainly go
* Reduction of repeated code is always good. However, in the case you linked you provided an overly general implementation, an I would argue that a `find_view_by_name` without templates is more readable and more descriptive.

I am aware that not having a style guide is not ideal. We have a [Developers Guide](https://github.com/simonfuhrmann/mve/wiki/MVE-Developers-Guide) where a Style section could be added (or a separate page, no strong opinion on that).

Let's just agree that we don't agree on the use of `auto`. I can agree with some the arguments, but it boils down to prioritization: I prefer readable code over "less refactoring overhead". While they argue your IDE could show you the type, I would argue that the IDE can also help with refactoring. So yea, I don't fully buy these arguments, and I'm sure you can find as many pro-`auto` articles than con-`auto` ones.

Having this said, I welcome `auto` in situations where the type is overly complex (although a `typedef` would also do, and we have lots of typedefs in the code base), and then there are these situations when you don't even want to bother with typedefs, like when using iterators.

PR #338, opened 2017-05-15
Removal of raw Scene and View constructors from public API

LinusVanElswijk:
This is a rework of PR #337.
Refactor of Scene and View class code to make it easier to read and maintain:

- Replaced index based loops, with foreach loops where possible
- Removed raw constructors from public API
- Removed commented out code
- Removed unused control flow parameters of internal functions
- Replaced pointer parameters by reference parameters where nullptr would be illegal as argument.

No functional changes.

### DIFF
--- a/libs/mve/scene.h
+++ b/libs/mve/scene.h
@@ -86,6 +86,9 @@ public:
     /** Returns key point memory usage. */
     std::size_t get_bundle_mem_usage (void);
 
+protected:
+    Scene (void);
+
 private:
     std::string basedir;
     ViewList views;
@@ -93,7 +96,6 @@ private:
     bool bundle_dirty;
 
 private:
-    Scene (void);
     void init_views (void);
 };
 
--- a/libs/mve/scene.h
+++ b/libs/mve/scene.h
@@ -38,13 +38,11 @@ public:
     typedef std::vector<View::Ptr> ViewList;
 
 public:
-    /** Constructs an unmanaged scene, which should not be copied. */
-    Scene (void);
-
-    /** Constructs a smart pointered scene. */
-    static Scene::Ptr create (void);
     /** Constructs and loads a scene from the given directory. */
     static Scene::Ptr create (std::string const& path);
+
+    Scene(const Scene&) = delete;
+    Scene& operator=(const Scene&) = delete;
 
     /** Loads the scene from the given directory. */
     void load_scene (std::string const& base_path);
@@ -95,6 +93,7 @@ private:
     bool bundle_dirty;
 
 private:
+    Scene (void);
     void init_views (void);
 };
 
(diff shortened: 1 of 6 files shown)


simonfuhrmann:
Can you please provide smaller PRs, ideally with one commit each? This is really hard to review, and smaller PRs have shorter round-trips.

LinusVanElswijk:
No problem, I'll work on that.

LinusVanElswijk:
Update: Changes are now restricted to the removal of unsafe constructors and factory functions from the public API of View and Scene:

- Removed raw constructors from View and Scene class API.
- Removed Scene::Create(void), which returns scenes in a weird and undocumented state. Directly initializing with Scene::Create(std::string) should be preferred and pointing to null should be preferred over  pointing to partially initialized scenes.

tests/mve/gtest_view.cc
```diff
@@ -8,176 +8,179 @@
```
LinusVanElswijk: The only real change in this file is the addition on line 12. Everything else is just replacement of a View with View::Ptr.

tests/mve/gtest_view.cc
```diff
@@ -8,176 +8,179 @@
...
-    view2.set_blob(blob, "blob");
-    EXPECT_TRUE(view2.is_dirty());
-
-    EXPECT_FALSE(view3.is_dirty());
-    EXPECT_FALSE(view3.get_meta_data().is_dirty);
-    view3.set_value("view.key", "value");
-    EXPECT_TRUE(view3.is_dirty());
-    EXPECT_TRUE(view3.get_meta_data().is_dirty);
+    mve::View::Ptr view1 = mve::View::create(),
+                   view2 = mve::View::create(),
```
simonfuhrmann: Please make three instructions for this.

libs/mve/scene.h
```diff
@@ -95,6 +93,7 @@ class Scene
     bool bundle_dirty;
 
 private:
+    Scene (void);
```
simonfuhrmann: It's customary to make this protected if the class is not final.

libs/mve/scene.cc
```diff
@@ -120,19 +96,18 @@ Scene::get_total_mem_usage (void)
 std::size_t
 Scene::get_view_mem_usage (void)
 {
-    std::size_t ret = 0;
-    for (std::size_t i = 0; i < this->views.size(); ++i)
-        if (this->views[i] != nullptr)
-            ret += this->views[i]->get_byte_size();
-    return ret;
+    return std::accumulate(views.begin(), views.end(), std::size_t{0},
```
LinusVanElswijk: It was not a conscient decision.
I got into the habit of using list initialization, because this form forbids narrowing conversions.

PR #342, opened 2017-05-19
Unit tests for mve::Scene.

LinusVanElswijk:
My current work on mve requires some extension to mve::Scene and without a good set of unit test available it is pretty hard to verify if anything broke. The unit test are set up to read like a specification and to provide clear messages on test failure. Please let me know if there are any errors in this specification.

This PR adds a function to remove directories, so that the unit tests can properly clean up after themselves. I did not test this function on windows.

At the moment two tests fail, because the bundles do not seem to save the pixel aspect ratio of cameras properly. They always reset to 1.0 on reload. Is this a bug or expected behavior?

```
...
[----------] Global test environment tear-down
[==========] 106 tests from 13 test cases ran. (265 ms total)
[  PASSED  ] 104 tests.
[  FAILED  ] 2 tests, listed below:
[  FAILED  ] SceneTest.when_save_is_called_on_a_scene_the_scene_on_disk_updates_accordingly
[  FAILED  ] SceneTest.when_save_bundle_is_called_on_a_scene_only_the_bundle_is_updated_on_disk

 2 FAILED TESTS
```

***Edit:*** Forgot to mention that bundle save/load is only partially checked: only cameras are checked for correctness, features are skipped.

### DIFF
--- a/tests/mve/gtest_scene.cc
+++ b/tests/mve/gtest_scene.cc
@@ -1,0 +1,615 @@
+#include <gtest/gtest.h>
+
+#include <cstdio>
+#include <iomanip>
+#include <sstream>
+#include <string>
+
+#include "mve/bundle_io.h"
+#include "mve/scene.h"
+#include "util/exception.h"
+#include "util/file_system.h"
+
+namespace {
+
+/* Cleanup mechanism. Unlinks files on destruction. */
+class OnScopeExit final
+{
+public:
+    ~OnScopeExit (void);
+    void unlink (std::string const& str);
+
+private:
+    void unlink_recursive (std::string const& str);
+
+private:
+    std::vector<std::string> paths;
+};
+
+std::string create_scene_on_disk (std::size_t const view_count,
+    mve::Bundle::Ptr const bundle, OnScopeExit* const clean_up);
+mve::Scene::Ptr scene_with_dirty_bundle_and_view (OnScopeExit* const clean_up);
+void make_dirty (mve::View::Ptr view);
+void make_a_clean_view_dirty (mve::Scene::Ptr scene);
+mve::Scene::ViewList load_views_from (std::string const& scene_path);
+mve::Bundle::Ptr load_bundle_from (std::string const& scene_path);
+mve::Bundle::Ptr make_bundle (std::size_t const camera_count);
+bool views_match (mve::Scene::ViewList const& lhs,
+    mve::Scene::ViewList const& rhs);
+bool bundle_cameras_match (mve::Bundle::Ptr lhs, mve::Bundle::ConstPtr rhs);
+
+} // anonymous namespace
+
+//== Test the initial state of a created scene =================================
+
+TEST(SceneTest, ACreatedSceneIsInitialyClean)
+{
+    OnScopeExit clean_up;
+
+    std::string scene_path = create_scene_on_disk(0, nullptr, &clean_up);
+    mve::Scene::Ptr scene = mve::Scene::create(scene_path);
+    ASSERT_TRUE(scene != nullptr);
+    EXPECT_FALSE(scene->is_dirty());
+}
+
+TEST(SceneTest, TheInitialPathOfACreatedSceneIsThePathItWasCreatedWith)
+{
+    OnScopeExit clean_up;
+
+    std::string scene_path = create_scene_on_disk(0, make_bundle(0), &clean_up);
+    mve::Scene::Ptr scene = mve::Scene::create(scene_path);
+    EXPECT_EQ(scene_path, scene->get_path());
+}
+
+TEST(SceneTest, TheInitialViewsOfACreatedSceneMatchWithThatSceneOnDisk)
+{
+    using ViewList = mve::Scene::ViewList;
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_without_views = mve::Scene::create(
+        create_scene_on_disk(0, make_bundle(5), &clean_up));
+    EXPECT_EQ(scene_without_views->get_views().size(), std::size_t(0));
+
+    mve::Scene::Ptr scene_with_views = mve::Scene::create(
+        create_scene_on_disk(73, make_bundle(23), &clean_up));
+    ViewList views_on_disk = load_views_from(scene_with_views->get_path());
+    EXPECT_TRUE(views_match(views_on_disk, scene_with_views->get_views()));
+}
+
+TEST(SceneTest, TheInitialBundleOfACreatedSceneMatchesWithThatSceneOnDisk)
+{
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_with_empty_bundle = mve::Scene::create(
+        create_scene_on_disk(0, make_bundle(0), &clean_up));
+    EXPECT_TRUE(bundle_cameras_match(
+        load_bundle_from(scene_with_empty_bundle->get_path()),
+        scene_with_empty_bundle->get_bundle()));
+
+    mve::Scene::Ptr scene_with_non_empty_bundle = mve::Scene::create(
+        create_scene_on_disk(3, make_bundle(23), &clean_up));
+    EXPECT_TRUE(bundle_cameras_match(
+        load_bundle_from(scene_with_non_empty_bundle->get_path()),
+        scene_with_non_empty_bundle->get_bundle()));
+}
+
+//== Creating a scene with missing files or directories ========================
+
+TEST(SceneTest, CreateSceneThrowsAnExceptionIfTheDirectoryDoesNotExist)
+{
+    std::string not_a_directory = std::tmpnam(nullptr);
+    EXPECT_THROW(mve::Scene::create(not_a_directory), util::Exception);
+}
+
+TEST(SceneTest, CreateSceneThrowsAnExceptionIfTheViewsSubdirectoryDoesNotExist)
+{
+    OnScopeExit clean_up;
+
+    std::string dir_with_no_views_subdir;
+    {
+        dir_with_no_views_subdir = std::tmpnam(nullptr);
+        util::fs::mkdir(dir_with_no_views_subdir.c_str());
+        clean_up.unlink(dir_with_no_views_subdir);
+        std::string bundle_file = util::fs::join_path(
+            dir_with_no_views_subdir, "synth_0.out");
+        mve::save_mve_bundle(make_bundle(0), bundle_file);
+    }
+    EXPECT_THROW(mve::Scene::create(dir_with_no_views_subdir), util::Exception);
+}
+
+TEST(SceneTest, CreatingASceneOnADirectoryWithNoBundleFileMakesGetBundleThrow)
+{
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_missing_bundle =
+        mve::Scene::create(create_scene_on_disk(0, nullptr, &clean_up));
+    EXPECT_THROW(scene_missing_bundle->get_bundle(), util::Exception);
+}
+
+//== Test loading into an existing scene =======================================
+
+TEST(SceneTest, WhenLoadIsCalledOnASceneItsPathUpdatesAccordingly)
+{
+    OnScopeExit clean_up;
+
+    std::string directory = create_scene_on_disk(0, make_bundle(3), &clean_up);
+    mve::Scene::Ptr scene = mve::Scene::create(
+        create_scene_on_disk(13, make_bundle(3), &clean_up));
+    scene->load_scene(directory);
+    EXPECT_EQ(directory, scene->get_path());
+}
+
+TEST(SceneTest, WhenLoadIsCalledOnASceneItsViewsUpdateAccordingly)
+{
+    using Views = mve::Scene::ViewList;
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene = mve::Scene::create(
+        create_scene_on_disk(13, make_bundle(3), &clean_up));
+    std::string path = create_scene_on_disk(9, make_bundle(4), &clean_up);
+    scene->load_scene(path);
+    Views views_from_disk = load_views_from(path);
+    EXPECT_TRUE(views_match(views_from_disk, scene->get_views()));
+}
+
+TEST(SceneTest, WhenLoadIsCalledOnASceneItsBundleUpdatesAccordingly)
+{
+    OnScopeExit clean_up;
+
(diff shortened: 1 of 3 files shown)


tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+#include "util/file_system.h"
+
+namespace {
+
+/* Cleanup mechanism. Unlinks files on destruction. */
+class OnScopeExit final
+{
+public:
+    ~OnScopeExit();
+    void unlink(std::string str);
```
simonfuhrmann: std::string const&

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+
+private:
+    void unlink_recursive(const std::string& str);
+
+    std::vector<std::string> paths;
```
simonfuhrmann: Open a new `private:` section to separate functions and variables.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+/**
+ * Creates a valid scene directory in the system's temporary directory.
+ * This directory has an empty "views" subdirectory and
+ * a bundle file with zero cameras and zero features.
+ * @param view_count Number of views the scene is created with.
+ * @param bundle Bundle to include within the scene or
+ *     nullptr to create a scene without a bundle file.
+ * @return The path to the created scene directory.
+ */
+std::string create_scene_on_disk(const std::size_t view_count,
```
simonfuhrmann: In MVE, just use up to 80 chars, and if the line needs to be wrapped, indent with four spaces. Don't align with brackets.
When declaring and defining functions, put a space before the bracket, like `int function (int value)`. If I'd have re-evaluate the style, I'd probably do away with this rule, but at this point let's just be consistent.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+    OnScopeExit clean_up;
+
+    std::string scene_path = create_scene_on_disk(0, nullptr, clean_up);
+    mve::Scene::Ptr scene = mve::Scene::create(scene_path);
+    ASSERT_TRUE(scene != nullptr);
+    EXPECT_FALSE(scene->is_dirty());
+}
+
+TEST(SceneTest,
+     the_initial_path_of_a_created_scene_is_the_path_it_was_created_with)
```
simonfuhrmann: Put the test names on the same line as TEST. Use CamelCase for the test names like in the other tests. If the test name doesn't fit, just make the name a bit shorter :)

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+
+void OnScopeExit::unlink(std::string str)
+{
+    this->paths.push_back(std::move(str));
+}
+
+void OnScopeExit::unlink_recursive(const std::string& path)
+{
+    namespace fs = util::fs;
+    if(fs::file_exists(path.c_str()))
```
simonfuhrmann: Space after for, if, while, etc.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+
+    if (lhs->get_cameras().size() != rhs->get_cameras().size() ||
+        lhs->get_features().size() != rhs->get_features().size())
+        return false;
+
+    for(std::size_t i = 0; i < lhs->get_cameras().size(); ++i)
+    {
+        const mve::CameraInfo& left_cam = lhs->get_cameras()[i];
+        const mve::CameraInfo& right_cam = rhs->get_cameras()[i];
+        if( !match(left_cam.flen, right_cam.flen)
```
simonfuhrmann: No space after bracket.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+        list.push_back(mve::View::create(file.get_absolute_name()));
+
+    return list;
+}
+
+mve::Bundle::Ptr
+load_bundle_directly_from(const std::string& scene_directory)
+{
+    std::string bundle_file =
+            util::fs::join_path(scene_directory, "synth_0.out");
```
simonfuhrmann: four spaces indent.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+
+std::string
+create_scene_on_disk(const std::size_t view_count,
+                     const mve::Bundle::Ptr bundle,
+                     OnScopeExit& on_scope_exit)
+{
+    namespace fs = util::fs;
+
+    std::string scene_directory { std::tmpnam(nullptr) };
+                scene_directory += "_test_scene";
```
simonfuhrmann: This is not properly indented

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+namespace {
+
+std::string
+create_scene_on_disk(const std::size_t view_count,
+                     const mve::Bundle::Ptr bundle,
+                     OnScopeExit& on_scope_exit)
+{
+    namespace fs = util::fs;
+
+    std::string scene_directory { std::tmpnam(nullptr) };
```
simonfuhrmann: Don't ever use {} for assignment. The {} operator is pretty evil. Consider this, what do you think this is doing?

    int a = 1, b = 10, c = 20;
    std::vector<int> test {&a, &c};

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+    if (lhs.size() != rhs.size())
+        return false;
+
+    for(mve::View::Ptr& left_view : lhs)
+    {
+        mve::Scene::ViewList::iterator found = std::find_if(
+                rhs.begin(),
+                rhs.end(),
+                [&left_view](mve::View::Ptr& right_view) {
+                    return left_view->get_id() == right_view->get_id();
```
LinusVanElswijk: Is it allowed to end a line earlier, for example to keep `rhs.begin()` and `rhs.end()` on the same line here?

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,751 @@
...
+}
+
+TEST(SceneTest,
+     reset_bundle_cleans_a_scene_if_only_its_bundle_is_dirty)
+{
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_with_dirty_bundle_and_clean_views = [&clean_up]() {
+        std::string path_to_scene =
+                create_scene_on_disk(10, make_bundle(3), clean_up);
```
LinusVanElswijk: How should this wrap.
As it is now, or like this
```
                                                                        | // column 80
std::string path_to_scene = create_scene_on_disk(10, make_bundle(3),    |
    clean_up);                                                          |
```

simonfuhrmann: Both is fine, but I prefer what you already have in code.

tests/mve/gtest_scene.cc
simonfuhrmann: If it improves readability considerably, you can. But here I'd argue it doesn't.

LinusVanElswijk:
Integrated the feedback into the unit test.

Unit test does not fail anymore. Figured out that bundle files only store a subset of the camera info and updated the unit tests take to that into account.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+#include "util/file_system.h"
+
+namespace {
+
+/* Cleanup mechanism. Unlinks files on destruction. */
+class OnScopeExit final
+{
+public:
+    ~OnScopeExit();
+    void unlink(const std::string& str);
```
simonfuhrmann: Space before bracket.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+
+private:
+    std::vector<std::string> paths;
+};
+
+/**
+ * Creates a valid scene directory in the system's temporary directory.
+ * This directory has an empty "views" subdirectory and
+ * a bundle file with zero cameras and zero features.
+ * @param view_count Number of views the scene is created with.
```
simonfuhrmann: Remove JS style comments. If your variables have useful names, this is generally not necessary.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+ * @param bundle Bundle to include within the scene or
+ *     nullptr to create a scene without a bundle file.
+ * @return The path to the created scene directory.
+ */
+std::string create_scene_on_disk (const std::size_t view_count,
+    const mve::Bundle::Ptr bundle, OnScopeExit& clean_up);
```
simonfuhrmann: Regarding function arguments: Either use const references, or pointers if the argument is modified. Let the types speak for themselves :)

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+TEST(SceneTest, TheInitialViewsOfACreatedSceneMatchWithThatSceneOnDisk)
+{
+    using ViewList = mve::Scene::ViewList;
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_without_views = [&clean_up]()
+    {
+        std::string path = create_scene_on_disk(0, make_bundle(5), clean_up);
+        return mve::Scene::create(path);
+    }();
```
simonfuhrmann: Why are you creating a lambda and calling it immediately? This is equivalent to writing:
````
std::string path = create_scene_on_disk(0, make_bundle(5), clean_up);
mve::Scene::Ptr scene_without_views = mve::Scene::create(path);
````
If you don't want the path temporary, just use a scope. Or just keep re-using the same path variable. Don't use lambdas if you can solve it with simpler tools, like just a scope.

File: tests/mve/gtest_scene.cc [hunk 16]
```diff
@@ -0,0 +1,715 @@
...
+    EXPECT_TRUE(bundle_cameras_match(
+        load_bundle_from(scene_with_non_empty_bundle->get_path()),
+        scene_with_non_empty_bundle->get_bundle()));
+}
+
+//== Creating a scene with missing files or directories ========================
+
+TEST(SceneTest, CreateSceneThrowsAnExceptionIfTheDirectoryDoesNotExist)
+{
+    std::string not_a_directory = tmpnam(nullptr);
```
simonfuhrmann: `std::tmpnam` here and everywhere else?

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+        view->save_view_as(view_directory_path);
+    }
+
+    if (bundle != nullptr)
+        mve::save_mve_bundle(bundle, bundle_file);
+
+    return scene_directory;
+}
+
+void make_dirty (mve::View::Ptr view)
```
simonfuhrmann: return types to their own line

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+            return false;
+    }
+
+    return true;
+}
+
+OnScopeExit::~OnScopeExit ()
+{
+    bool exception_happened = false;
+    for (std::string path : this->paths)
```
simonfuhrmann: std::string const&

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,715 @@
...
+            exception_happened = true;
+            std::cout <<  "Exception during cleanup!" << std::endl;
+        }
+    }
+    if (exception_happened)
+        throw std::runtime_error("Exception during file cleanup.");
+}
+
+void
+OnScopeExit::unlink (const std::string& str)
```
simonfuhrmann: In MVE, we write `var const&` instead of `const var&`. If you wonder why, it's because `const` ALWAYS applies to the left. Consider `var const*` and `var* const` (either var is const, or the pointer is const). Consider `void my_member (void) const` (i.e. const applies to the function on its left). There is only one exception where const applies to the right (when there is no left), and there is really no need to ever use this exception.

nmoehrle: +1

tests/mve/gtest_scene.cc
LinusVanElswijk: Not sure what you mean here.
Most test cases use create_scene_on_disk to make a populated scene directory to test with, but this particular test needs a path to a directory that does not exist.

simonfuhrmann: `tmpnam(...)` -> `std::tmpnam(...)`

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,618 @@
...
+{
+    OnScopeExit clean_up;
+
+    mve::Scene::Ptr scene_with_dirty_bundle = [&clean_up]()
+    {
+        std::string path = create_scene_on_disk(13, make_bundle(15), &clean_up);
+        mve::Scene::Ptr scene = mve::Scene::create(path);
+        scene->set_bundle(make_bundle(0));
+        return scene;
+    }();
```
LinusVanElswijk: Need to remove the immediately invoked lambda.

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,618 @@
...
+    std::string bundle_file = fs::join_path(scene_directory, "synth_0.out");
+    std::string views_directory = fs::join_path(scene_directory, "views");
+
+    fs::mkdir(scene_directory.c_str());
+    on_scope_exit->unlink(scene_directory); // Schedules for cleanup.
+    fs::mkdir(views_directory.c_str());
+
+    for (std::size_t i = 0; i < view_count; ++i)
+    {
+        const std::string view_directory_path = [&i, &views_directory]()
```
LinusVanElswijk: const std::string -> std::string const

tests/mve/gtest_scene.cc
```diff
@@ -0,0 +1,618 @@
...
+
+mve::Scene::ViewList
+load_views_from (std::string const& scene_directory)
+{
+    std::string view_directory = util::fs::join_path(scene_directory, "views");
+    util::fs::Directory directory(view_directory);
+    mve::Scene::ViewList list;
+    list.reserve(directory.size());
+
+    for (const util::fs::File& file : directory)
```
LinusVanElswijk: const util::fs::File& -> util::fs::File& const

LinusVanElswijk:
Code should now be up to standard.

simonfuhrmann:
Thanks for the contribution.

PR #349, opened 2017-06-04
Add essential -mxxx build flag for successfull gcc6/linux compilation.

bartoszek:
Without -msse4.2 gcc7 complains about `target specific option mismatch `

### DIFF
--- a/libs/sfm/Makefile
+++ b/libs/sfm/Makefile
@@ -2,7 +2,7 @@ MVE_ROOT := ../..
 TARGET := libmve_sfm.a
 include ${MVE_ROOT}/Makefile.inc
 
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
I am wondering if this will break compilation on older compilers. I don't think we use SSE4 anywhere, do we?

flanggut:
SSE4 is ~10 yeas old by now - I think if your compiler supports c++11 it should also support SSE4.
I also tested this with LLVM without any problems. I would, however, directly add the option to the `CXXINTRINSICS` in`Makefile.inc`.

simonfuhrmann:
Moved to `Makefile.inc`. Please check if `-msse4` is sufficient instead of `-msse4.2`.

PR #366, opened 2017-07-20
OGL: report program linking errors

seepa:
Similar to the shader compilation error reporting, we should also show the errors when linking shaders. This PR implements this by retrieving the program info log in case the linking failed after the ``glLinkProgram`` call.

### DIFF
--- a/libs/ogl/shader_program.h
+++ b/libs/ogl/shader_program.h
@@ -300,17 +300,6 @@ ShaderProgram::unbind (void) const
     check_gl_error();
 }
 
-inline void
-ShaderProgram::ensure_linked (void)
-{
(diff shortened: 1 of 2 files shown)


libs/ogl/shader_program.h
```diff
@@ -307,6 +308,17 @@ ShaderProgram::ensure_linked (void)
     {
         glLinkProgram(this->prog_id);
         check_gl_error();
+        if (this->get_program_property(GL_LINK_STATUS) == GL_FALSE)
```
simonfuhrmann: Please move this function to the .cc file.

seepa:
kk, done.

Issue #359, opened 2017-06-29
error: inlining failed in call to always_inline ‘long long int _mm_popcnt_u64(long long unsigned int)’: target specific option mismatch


ckorn:
Hi,

on Ubuntu 17.04 there is this compilation error:
```
g++ -g -O2 -fdebug-prefix-map=/home/korn/packages/mve/mve-0.0.0+git20170629=. -fstack-protector-strong -Wformat -Werror=format-security -pthread -I../../libs -fopenmp -Wdate-time -D_FORTIFY_SOURCE=2 -c -o pose_p3p.o pose_p3p.cc
In file included from ../../libs/sfm/cascade_hashing.h:20:0,
                 from cascade_hashing.cc:14:
/usr/lib/gcc/x86_64-linux-gnu/6/include/popcntintrin.h: In member function ‘void sfm::CascadeHashing::collect_features_from_buckets(std::vector<std::vector<unsigned int> >*, size_t, std::vector<bool>*, const BucketGroupsBuckets&, const BucketGroupsFeatures&, const uint64_t*, const std::vector<long unsigned int, std::allocator<long unsigned int> >&) const [with int DIMHASH = 128]’:
/usr/lib/gcc/x86_64-linux-gnu/6/include/popcntintrin.h:42:1: error: inlining failed in call to always_inline ‘long long int _mm_popcnt_u64(long long unsigned int)’: target specific option mismatch
 _mm_popcnt_u64 (unsigned long long __X)
 ^~~~~~~~~~~~~~
In file included from cascade_hashing.cc:14:0:
../../libs/sfm/cascade_hashing.h:452:47: note: called from here
                 hamming_dist += _mm_popcnt_u64(comp_hash_data1[k] ^ ptr2[k]);
                                 ~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from ../../libs/sfm/cascade_hashing.h:20:0,
                 from cascade_hashing.cc:14:
/usr/lib/gcc/x86_64-linux-gnu/6/include/popcntintrin.h:42:1: error: inlining failed in call to always_inline ‘long long int _mm_popcnt_u64(long long unsigned int)’: target specific option mismatch
```

flanggut:
The compile command doesn't activate the popcount intrinsic. What build system do you use?

ckorn:
This is with the make build system.

Here, should have provided the entire build log.
[...]

PR #372, opened 2017-09-19
SfM: Changed popcount implementation to use std::bitset::count()

LinusVanElswijk:
Related to issue #359.

### DIFF
--- a/libs/math/functions.h
+++ b/libs/math/functions.h
@@ -10,7 +10,13 @@
 #ifndef MATH_FUNCTIONS_HEADER
 #define MATH_FUNCTIONS_HEADER
 
+#include <bitset>
 #include <cmath>
+#include <cinttypes>
+
+#ifdef _MSC_VER
+#   include <intrin.h>
+#endif
 
 #include "math/defines.h"
 #include "math/vector.h"
@@ -151,6 +157,42 @@ sinc (T const& x)
 {
     return (x == T(0) ? T(1) : std::sin(x) / x);
 }
+
+/* ----------------------------- Popcount ------------------------- */
+
+/** Returns the number of one bits of an integer. */
+template <typename T>
+std::size_t constexpr
+popcount (T const x)
+{
+    static_assert(std::is_integral<T>::value && std::is_unsigned<T>::value,
+        "T must be an unsigned integral type.");
+    return std::bitset<sizeof(T) * 8>(x).count();
+}
+
+#ifdef _MSC_VER // MSVS fails to optimize std::bitset::count
+std::size_t inline
+popcount (std::uint32_t const x)
+{
+#   if !(__amd64__ || __x86_64__ || _WIN64 || _M_X64)
+    __pragma (message("Warning: CPU support for _mm_popcnt_u32 is assumed, "
+        "but could not be verified."))
+#   endif
+    return static_cast<std::size_t>(_mm_popcnt_u32(x));
+}
+
+std::size_t inline
+popcount (std::uint64_t const x)
+{
+#   if __amd64__ || __x86_64__ || _WIN64 || _M_X64
+    return static_cast<std::size_t >(_mm_popcnt_u64(x));
+#   else
+    std::uint32_t const low_bits = static_cast<std::uint32_t>(x);
+    std::uint32_t const high_bits = static_cast<std::uint32_t>(x >> 32);
+    return popcount(low_bits) + popcount(high_bits);
+#   endif
+}
+#endif
 
 /* ------------------------------ Misc ---------------------------- */
 
--- a/libs/sfm/cascade_hashing.h
+++ b/libs/sfm/cascade_hashing.h
@@ -14,12 +14,7 @@
 #include <iostream>
 #include <random>
 
-#ifdef _MSC_VER
-#   include <nmmintrin.h>
-#else
-#   include <popcntintrin.h>
-#endif
-
+#include "math/functions.h"
 #include "math/vector.h"
 #include "sfm/defines.h"
 #include "sfm/exhaustive_matching.h"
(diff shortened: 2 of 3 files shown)


simonfuhrmann:
Is this negatively affecting performance?

pmoulon:
I would be curious too to know.

- I think it depends of the arch and of the compiler flag.
- I don't know how Visual Studio is dealing with bitset::count implementation by default.

LinusVanElswijk:
I did some more experiments.

I assume that -O2 or -O3 is used.
I also assume that newer compilers versions, produce code that is at least as fast as older versions.

With clang (3.5) and gcc (5.1), the resulting machine code is equivalent to using the popcount intrinsic.

I did not test MSVS before. Sadly it seems like is MSVS is unable to optimize away the std::bitset and the assembly looks horrible.

Since the MSVS assembly turns out to be so much worse, I need to either need to identify the code as cold code or provide a MSVS specific implementation.

simonfuhrmann:
Even if it is slower, what relative numbers are we talking? If it's 1% slower, then I don't care :)

simonfuhrmann:
Any update on this? If performance is not significantly impacted, I'm going to pull this.

pmoulon:
I made a test on windows (with the OpenMVG Hamming distance metric) and definitively bitset popcount is not accelerated by default in Visual Studio. If you want to be as fast as you can you have to use the CPU intrinsics. Only GCC and CLANG are doing the right job to switch to the optimized settings.

- My experiments show a 6x speed improvement if I call the CPU intrinsic on VStudio compare to the `bitset<>.count()`

simonfuhrmann:
Thanks Pierre. Maybe we should put an ifdef around this line of code, call the new function from this PR if the intrinsic is not available, and keep using the intrinsic otherwise. @LinusVanElswijk, can you add this ifdef?

andre-schulz:
@pmoulon which Visual Studio version did you test?

pmoulon:
VS2015

LinusVanElswijk:
@pmoulon thank you for doing the test.
I will add the ifdefs statements and update the PR today.

libs/math/functions.h
```diff
@@ -152,6 +158,42 @@ sinc (T const& x)
...
+/** Returns the number of one bits of an integer. */
+template <typename T>
+std::size_t constexpr
+popcount (T const x)
+{
+    static_assert(std::is_integral<T>::value && std::is_unsigned<T>::value,
+        "T must be an unsigned integral type.");
+    return std::bitset<sizeof(T) * 8>(x).count();
+}
+
```
LinusVanElswijk: The following conditional compilation instructions are based on the [openMVG hamming distance implementation](https://github.com/openMVG/openMVG/blob/68def23c559248bbfb64616526ebbfec341c8c1a/src/openMVG/matching/metric_hamming.hpp#L61).

libs/math/functions.h
```diff
@@ -152,6 +158,42 @@ sinc (T const& x)
...
+#ifdef _MSC_VER // MSVS fails to optimize std::bitset::count
+std::size_t inline
+popcount (std::uint32_t const x)
+{
+#   if !(__amd64__ || __x86_64__ || _WIN64 || _M_X64)
+    __pragma (message("Warning: CPU support for _mm_popcnt_u32 is assumed, "
+        "but could not be verified."))
+#   endif
```
LinusVanElswijk: Not sure how to check for CPU support here and settled on just assuming availability and showing a warning for now.

libs/math/functions.h
```diff
@@ -152,6 +158,42 @@ sinc (T const& x)
...
 
+/* ----------------------------- Popcount ------------------------- */
+
+/** Returns the number of one bits of an integer. */
+template <typename T>
+std::size_t constexpr
+popcount (T const x)
+{
+    static_assert(std::is_integral<T>::value && std::is_unsigned<T>::value,
+        "T must be an unsigned integral type.");
```
LinusVanElswijk: Added requirement for T to be unsigned.
Without this, MSVS would silently use the unoptimized bitset::count for all signed integers.

I tried adding overloading for the signed types on MSVS, but this increased the line count a bit too much for little added benefit.

libs/math/functions.h
```diff
@@ -152,6 +158,42 @@ sinc (T const& x)
...
+
+std::size_t inline
+popcount (std::uint64_t const x)
+{
+#   if __amd64__ || __x86_64__ || _WIN64 || _M_X64
+    return static_cast<std::size_t >(_mm_popcnt_u64(x));
+#   else
+    std::uint32_t const low_bits = static_cast<std::uint32_t>(x);
+    std::uint32_t const high_bits = static_cast<std::uint32_t>(x >> 32);
+    return popcount(low_bits) + popcount(high_bits);
```
LinusVanElswijk: Help MSVS count to 64

libs/math/functions.h
pmoulon: Did you got an error or a warning?
I don't have any trouble on VS2015.

I got complains about `std::uint32_t const low_bits ` if I let the `constexpr` since you can't create variable.

LinusVanElswijk: I got an error just for calling _mm_popcount in a constexpr.
Not sure what causes the difference in behavior.

[Test case](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(j:1,source:'%23include+%3Ccstddef%3E%0A%23include+%3Ccinttypes%3E%0A%23include+%3Cintrin.h%3E%0A%0Astd::size_t+constexpr%0Apopcount+(std::uint32_t+const+x)%0A%7B%0A++++return+static_cast%3Cstd::size_t%3E(_mm_popcnt_u32(x))%3B%0A%7D%0A%0A//+need+a+non-inline+function+using+popcount%0A//+to+get+the+assembly%0Astd::size_t+f(std::uint32_t+x)+%7B+return+popcount(x)%3B+%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:38.56407698973815,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:compiler,i:(compiler:cl19_32,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'1',intel:'0',trim:'0'),libs:!(),options:'',source:1),l:'5',n:'0',o:'x86+CL+19+2017+RTW+(Editor+%231,+Compiler+%231)',t:'0')),k:24.634306689968408,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1),l:'5',n:'0',o:'%231+with+x86+CL+19+2017+RTW',t:'0')),k:36.801616320293434,l:'4',n:'0',o:'',s:0,t:'0')),l:'2',n:'0',o:'',t:'0')),version:4)

PR #393, opened 2018-01-08
SFM: Fix conversion of radial distortion in bundler

flanggut:
The image undistortion expects distortion parameters in original camera coordinates and handles the focal length itself. So this conversion here is actually a bug, and we create wrong undistorted images later on.

I don't remember why we did this, was there ever a reason to do this multiplication?


simonfuhrmann:
I think the camera parameters in the bundle are not relevant, and not read by MVE. I think the multiplication was done so that the format is inline with what other bundlers output. I'm fine with changing it though.

flanggut:
We should definitely change this. `sfmrecon` actually uses these parameters for undistortion, so we've basically created wrong undist images for a while now...

PR #398, opened 2018-02-05
SFM: Always consider radial distortion

flanggut:
- Add an undistortion function for image features and use corrected features for track triangulation.
- If possible restore initial outlier tracks for viewports after intrinsics have been optimized.

### DIFF
--- a/libs/sfm/bundler_common.cc
+++ b/libs/sfm/bundler_common.cc
@@ -266,5 +266,56 @@ load_survey_from_file (std::string const& filename,
     in.close();
 }
 
+/* ---------------------- Feature undistortion -------------------- */
+
+namespace {
+    double distort_squared_radius (double const r2, double const k1,
+        double const k2)
+    {
+        // Compute the distorted squared radius:
+        // (radius * distortion_coeff)^2 = r^2 * coeff^2
+        double coeff = 1.0 + r2 * k1 + r2 * r2 * k2;
+        return r2 * coeff * coeff;
+    }
+
+    double solve_undistorted_squared_radius (double const r2,
+        double const k1, double const k2)
+    {
+        // We use bisection as we can easily find a pretty good initial guess,
+        // and we don't want to compute all roots for a 5th degree polynomial.
+
+        // Guess initial interval upper and lower bound
+        double lbound = r2, ubound = r2;
+        while (distort_squared_radius(lbound, k1, k2) > r2)
+            lbound /= 1.05;
+        while (distort_squared_radius(ubound, k1, k2) < r2)
+            ubound *= 1.05;
+
+        // Perform a bisection until epsilon accuracy is reached
+        while (std::numeric_limits<double>::epsilon() < ubound - lbound)
+        {
+            double const mid = 0.5 * (lbound + ubound);
+            if (distort_squared_radius(mid, k1, k2) > r2)
+                ubound = mid;
+            else
+                lbound = mid;
+        }
+        // Return center of interval
+        return 0.5 * (lbound + ubound);
+    }
+}
+
+math::Vec2f
+undistort_feature (math::Vec2f const& f, double const k1, double const k2,
+    float const focal_length)
+{
+    // Convert to camera coords
+    double const r2 = f.square_norm() / MATH_POW2(focal_length);
+    double scale = 1.0;
+    if (r2 > 0.0)
+        scale = std::sqrt(solve_undistorted_squared_radius(r2, k1, k2) / r2);
+    return f * scale;
+}
+
 SFM_BUNDLER_NAMESPACE_END
 SFM_NAMESPACE_END
--- a/libs/sfm/bundler_incremental.cc
+++ b/libs/sfm/bundler_incremental.cc
@@ -152,8 +152,8 @@ Incremental::reconstruct_next_view (int view_id)
 
     /*
      * Remove outliers from tracks and tracks from viewport.
-     * TODO: Once single cam BA has been performed and parameters for this
-     * camera are optimized, evaluate outlier tracks and try to restore them.
+     * Once single cam BA has been performed and parameters for this
+     * camera are optimized, tracks are evaluated again and restored.
      */
     for (std::size_t i = 0; i < ransac_result.inliers.size(); ++i)
         track_ids[ransac_result.inliers[i]] = -1;
@@ -163,6 +163,8 @@ Incremental::reconstruct_next_view (int view_id)
             continue;
         this->tracks->at(track_ids[i]).remove_view(view_id);
         this->viewports->at(view_id).track_ids[feature_ids[i]] = -1;
+        this->viewports->at(view_id).backup_tracks.emplace(
+            feature_ids[i], track_ids[i]);
     }
     track_ids.clear();
     feature_ids.clear();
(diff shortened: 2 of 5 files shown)


libs/sfm/bundler_incremental.cc
```diff
@@ -188,6 +190,41 @@ Incremental::reconstruct_next_view (int view_id)
     return true;
 }
 
+void
+Incremental::try_restore_tracks_for_views (void)
+{
+  for (std::size_t i = 0; i < this->viewports->size(); ++i)
```
nmoehrle: Did you use two spaces for indentation on purpose?

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
...
+    {
+        // Guess initial interval upper and lower bound
+        double lbound = r2, ubound = r2;
+        while (distort_radius(lbound, params) > r2)
+            lbound /= 1.05;
+        while (distort_radius(ubound, params) < r2 )
+            ubound *= 1.05;
+
+        // Perform a bisection until epsilon accuracy is reached
+        double const epsilon = 1e-10;
```
simonfuhrmann: You could use std::numeric_limits<double>::epsilon(), which is close to what you have, 1e-9.

libs/sfm/bundler_incremental.cc
```diff
@@ -188,6 +190,41 @@ Incremental::reconstruct_next_view (int view_id)
...
+            || viewport.track_ids[feature_id] >= 0)
+            continue;
+        math::Vec3f const& pos3d = this->tracks->at(track_id).pos;
+        math::Vec2f const& pos2d = undistort_feature(
+            features.positions[feature_id], viewport.radial_distortion,
+            viewport.focal_length);
+
+        math::Vec4d const pos3dh(pos3d[0], pos3d[1], pos3d[2], 1.0);
+        math::Matrix<double, 3, 4> P;
+        viewport.pose.fill_p_matrix(&P);
```
nmoehrle: I would help the compiler here and move the P matrix out of the loop :-)

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
     in.close();
 }
 
+/* ---------------------- Feature undistortion -------------------- */
+
+namespace {
+    double distort_radius (double const r2, float const* params)
+    {
+        double coeff = 1.0 + r2 * params[0] + r2 * r2 * params[1];
```
simonfuhrmann: Can you add a comment what you're computing in this function? I don't understand it. According to this formula, we return `r^2 * (1 + k0 r^2 + k2 r^4)^2` which contains the radius to the power of 8?!?

libs/sfm/bundler_incremental.cc
flanggut: whoops :D

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
...
+        return r2 * coeff * coeff;
+    }
+
+    double solve_undistorted_radius (double const r2, float const* params)
+    {
+        // Guess initial interval upper and lower bound
+        double lbound = r2, ubound = r2;
+        while (distort_radius(lbound, params) > r2)
+            lbound /= 1.05;
+        while (distort_radius(ubound, params) < r2 )
```
simonfuhrmann: remove space before bracket here, and two more times below.

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
...
+namespace {
+    double distort_radius (double const r2, float const* params)
+    {
+        double coeff = 1.0 + r2 * params[0] + r2 * r2 * params[1];
+        return r2 * coeff * coeff;
+    }
+
+    double solve_undistorted_radius (double const r2, float const* params)
+    {
+        // Guess initial interval upper and lower bound
```
simonfuhrmann: Is there a formula to compute this close form, instead of bisecting it?

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
...
+}
+
+math::Vec2f
+undistort_feature (math::Vec2f const& f, float const* params,
+    float const focal_length)
+{
+    // Convert to camera coords
+    double const r2 = f.square_norm() / MATH_POW2(focal_length);
+    double const undist_radius = solve_undistorted_radius(r2, params);
+    double const scale = (r2 == 0) ? 1. : std::sqrt(undist_radius / r2);
```
simonfuhrmann:
```
double scale = 1.0;
if (r2 != 0.0)
{
    double const undist_radius = ...
    scale = ...
}
```

This way we avoid computing undist radius if `r2 == 0`.

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
...
+    math::Vec2f res = f * scale;
```
simonfuhrmann: `return f * scale;`

libs/sfm/bundler_incremental.cc
```diff
@@ -188,6 +190,41 @@ Incremental::reconstruct_next_view (int view_id)
     return true;
 }
 
+void
+Incremental::try_restore_tracks_for_views (void)
+{
+  for (std::size_t i = 0; i < this->viewports->size(); ++i)
+  {
+    Viewport & viewport = this->viewports->at(i);
```
simonfuhrmann: No space before `&`

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,50 @@ load_survey_from_file (std::string const& filename,
     in.close();
 }
 
+/* ---------------------- Feature undistortion -------------------- */
+
+namespace {
+    double distort_radius (double const r2, float const* params)
```
simonfuhrmann: Should this function explicitly take the undistortion parameters, i.e., `k0` and `k2`, or whatever it is? This way it's easier to understand and you may implicitly convert those to doubles (i.e., take `double k0` and `double k2` as parameters).

flanggut: changed

libs/sfm/bundler_common.cc
flanggut: It simply computes the squared distorted radius. I added a comment to clarify this.

libs/sfm/bundler_common.cc
flanggut: added comment

flanggut: We would have to find and test all roots of a large polynomial, which is not practical.

libs/sfm/bundler_common.cc
```diff
@@ -266,5 +266,56 @@ load_survey_from_file (std::string const& filename,
...
+
+        // Guess initial interval upper and lower bound
+        double lbound = r2, ubound = r2;
+        while (distort_squared_radius(lbound, k1, k2) > r2)
+            lbound /= 1.05;
+        while (distort_squared_radius(ubound, k1, k2) < r2)
+            ubound *= 1.05;
+
+        // Perform a bisection until epsilon accuracy is reached
+        while (std::numeric_limits<double>::epsilon() < ubound - lbound)
```
simonfuhrmann: you need std::abs()?

flanggut: The upper bound is always bigger than the lower bound.

PR #404, opened 2018-02-22
Outputting dense 2D<-->3D correspondence data

IdoMa:
I'd be happy to contribute the "output dense correspondence" feature to MVE. Please have a look at this PR.

### DIFF
--- a/apps/scene2pset/scene2pset.cc
+++ b/apps/scene2pset/scene2pset.cc
@@ -100,17 +100,17 @@ save_correspondence_data(const CorrespondenceData& corr_data, const AppSettings&
     corr_data_file << "x, y\n";
     for (unsigned int i = 0; i < corr_data.data.size(); ++i)
     {
-        corr_data_file << std::to_string(corr_data.data[i][0]) << ", " <<
-            std::to_string(corr_data.data[i][1]) << "\n" ;
+        corr_data_file << corr_data.data[i][0] << ", " <<
+            corr_data.data[i][1] << "\n" ;
     }
 
     corr_metadata_file << "View_ID, Width, Height, First_Vertex_Index\n";
     for (unsigned int i = 0; i < corr_data.metadata.size(); ++i)
     {
-        corr_metadata_file << std::to_string(corr_data.metadata[i].view_id) << ", " <<
-            std::to_string(corr_data.metadata[i].width) << ", " <<
-            std::to_string(corr_data.metadata[i].height) << ", " <<
-            std::to_string(corr_data.metadata[i].first_idx) << "\n" ;
+        corr_metadata_file << corr_data.metadata[i].view_id << ", " <<
+            corr_data.metadata[i].width << ", " <<
+            corr_data.metadata[i].height << ", " <<
+            corr_data.metadata[i].first_idx << "\n" ;
     }
 
     corr_data_file.close();
--- a/apps/scene2pset/scene2pset.cc
+++ b/apps/scene2pset/scene2pset.cc
@@ -8,6 +8,7 @@
  */
 
 #include <iostream>
+#include <fstream>
 #include <stdexcept>
 #include <algorithm>
 #include <vector>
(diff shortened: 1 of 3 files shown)


libs/mve/depthmap.h
```diff
@@ -71,6 +71,8 @@ MVE_NAMESPACE_END
 MVE_NAMESPACE_BEGIN
 MVE_GEOM_NAMESPACE_BEGIN
 
+const float dd_factor_default = 5.0f;
```
simonfuhrmann: `DD_FACTOR_DEFAULT`

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    for (unsigned int i=0 ; i<pixel_amount ; ++i)
+        if (vertex_ids.at(i) != MATH_MAX_UINT)
+            curr_view_data[vertex_ids.at(i)] = math::Vec2ui(i%width,int(i/width));
+    corr_data.data.insert(corr_data.data.end(), curr_view_data.begin(), curr_view_data.end());
+}
+
+void
+save_correspondence_data(const CorrespondenceData& corr_data, const AppSettings& conf)
+{
+    std::string corr_data_fname     = conf.outmesh + "_correspondence-data.csv";
```
simonfuhrmann: no extra indents within the line.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    std::string corr_metadata_fname = conf.outmesh + "_correspondence-metadata.csv";
+
+    std::ofstream corr_data_file;
```
simonfuhrmann: open file here by passing the filename.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    std::ofstream corr_metadata_file;
+
```
simonfuhrmann: Check if open was successful
`if (!corr_data_file.good()) ...`

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    corr_data_file.open(corr_data_fname);
+    corr_metadata_file.open(corr_metadata_fname);
+
+    corr_data_file << "x, y\n";
+    for (unsigned int i=0 ; i<corr_data.data.size() ; ++i)
```
simonfuhrmann: for loop style example:
`for (int x = 0; x < size; ++x)`
Here and below.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    {
+        corr_data_file << std::to_string(corr_data.data[i][0]) + ", " +
+                          std::to_string(corr_data.data[i][1]) + "\n" ;
```
simonfuhrmann: Indent new lines with 4 spaces here and below.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+    std::ofstream corr_data_file;
+    std::ofstream corr_metadata_file;
+
+    corr_data_file.open(corr_data_fname);
+    corr_metadata_file.open(corr_metadata_fname);
+
+    corr_data_file << "x, y\n";
+    for (unsigned int i=0 ; i<corr_data.data.size() ; ++i)
+    {
+        corr_data_file << std::to_string(corr_data.data[i][0]) + ", " +
```
simonfuhrmann: don't use `+`, use `<<` for efficiency here and below.

apps/scene2pset/scene2pset.cc
```diff
@@ -104,13 +173,13 @@ main (int argc, char** argv)
     args.add_option('p', "poisson-normals", false, "Scale normals according to confidence");
     args.add_option('S', "scale-factor", true, "Factor for computing scale values [2.5]");
     args.add_option('F', "fssr", true, "FSSR output, sets -nsc and -di with scale ARG");
+    args.add_option('C', "correspondence", false, "Output correspondence data (only in the absence of -m and -b)");
```
simonfuhrmann: Keep lines to 80 chars.
`Output correspondences (in absence of -m and -b only)`

apps/scene2pset/scene2pset.cc
```diff
@@ -142,6 +211,7 @@ main (int argc, char** argv)
                     : "undist-L" + util::string::get<int>(scale);
                 break;
             }
+            case 'C': conf.output_correspondence = true; break;
```
simonfuhrmann: move above `case 'F'`

apps/scene2pset/scene2pset.cc
```diff
@@ -230,7 +304,13 @@ main (int argc, char** argv)
 
         /* Triangulate depth map. */
         mve::TriangleMesh::Ptr mesh;
-        mesh = mve::geom::depthmap_triangulate(dm, ci, cam);
+
+        mve::Image<unsigned int> vertex_ids;
+        if (conf.output_correspondence && conf.aabb.empty() && conf.mask.empty())
```
simonfuhrmann: Is this `if` necessary? Essentially, obtaining `vertex_ids` is free, no?

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
     float min_valid_fraction = 0.0f;
     float scale_factor = 2.5f; /* "Radius" of MVS patch (usually 5x5). */
     std::vector<int> ids;
+    bool output_correspondence = false;
 };
 
+struct CorrespondenceViewMetadata
+{
+    unsigned int view_id; 
+    unsigned int width;  // After downsampling.
```
simonfuhrmann: Remove the comments. Either it's clear, or unnecessary, or encode it in the variable name, e.g., `scaled_width`.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+};
+
+struct CorrespondenceData
+{
+    std::vector<math::Vec2ui> data;
+    std::vector<CorrespondenceViewMetadata> metadata;
+};
+
+void
+append_correspondence_data_from_view(CorrespondenceData& corr_data,
```
simonfuhrmann: Don't pass non-const references. If you want to modify a variable, pass it as last argument as pointer. MVE may not do this consistently, but it's good practice. This way it's explicit at the caller site that a variable gets modified.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
 };
 
+struct CorrespondenceViewMetadata
+{
+    unsigned int view_id; 
+    unsigned int width;  // After downsampling. 
+    unsigned int height; // After downsampling.
+    unsigned long first_idx;
+
+    CorrespondenceViewMetadata (unsigned int view_id, unsigned int width, unsigned int height, unsigned long first_idx):
```
simonfuhrmann: I'd remove the constructor here. It makes the code below a bit more verbose, but also more clear. And structs shouldn't have logic, even simple constructor logic.

apps/scene2pset/scene2pset.cc
```diff
@@ -43,8 +44,76 @@ struct AppSettings
...
+};
+
+void
+append_correspondence_data_from_view(CorrespondenceData& corr_data, 
+    const mve::Image<unsigned int>& vertex_ids, const int view_id)
+{   
+    const unsigned int width = vertex_ids.width();
+    corr_data.metadata.emplace_back(view_id, width, vertex_ids.height(), corr_data.data.size());
+    
+    unsigned int pixel_amount = vertex_ids.get_pixel_amount();
```
simonfuhrmann: `const` for these.

apps/scene2pset/scene2pset.cc
IdoMa: Essentially free indeed. However, `vertex_ids` will not be used later in the code if this condition doesn't hold.

IdoMa:
It is indeed possible to stream the correspondence data right away. I chose not to do so in order to comply with the existing layout - as the point-set is aggregated while iterating through the views, and saved to the disk only at the end.

apps/scene2pset/scene2pset.cc
```diff
@@ -66,48 +62,55 @@ struct CorrespondenceData
...
+    }
 
     corr_data_file << "x, y\n";
-    for (unsigned int i=0 ; i<corr_data.data.size() ; ++i)
+    for (unsigned int i = 0; i < corr_data.data.size(); ++i)
     {
-        corr_data_file << std::to_string(corr_data.data[i][0]) + ", " +
-                          std::to_string(corr_data.data[i][1]) + "\n" ;
+        corr_data_file << std::to_string(corr_data.data[i][0]) << ", " <<
+            std::to_string(corr_data.data[i][1]) << "\n" ;
```
simonfuhrmann: No need to convert to string if you use `<<`.

apps/scene2pset/scene2pset.cc
```diff
@@ -66,48 +62,55 @@ struct CorrespondenceData
...
     corr_metadata_file << "View_ID, Width, Height, First_Vertex_Index\n";
-    for (unsigned int i=0 ; i<corr_data.metadata.size() ; ++i)
+    for (unsigned int i = 0; i < corr_data.metadata.size(); ++i)
     {
-        corr_metadata_file << std::to_string(corr_data.metadata[i].view_id  ) + ", " +
-                              std::to_string(corr_data.metadata[i].width)     + ", " +
-                              std::to_string(corr_data.metadata[i].height)    + ", " +
-                              std::to_string(corr_data.metadata[i].first_idx) + "\n" ;
+        corr_metadata_file << std::to_string(corr_data.metadata[i].view_id) << ", " <<
+            std::to_string(corr_data.metadata[i].width) << ", " <<
```
simonfuhrmann: Here too.

IdoMa:
:)

PR #411, opened 2018-03-23
MVE: remove dead lines from image_io; remove C struct keyword

magcks:


libs/mve/image_io.cc
```diff
@@ -594,17 +594,7 @@ save_jpg_file (ByteImage::ConstPtr image, std::string const& filename, int quali
     cinfo.image_width = image->width();
     cinfo.image_height = image->height();
     cinfo.input_components = image->channels();
-    switch (image->channels())
```
simonfuhrmann: This is not dead code.

flanggut:
I just had a quick glance but I think the number of channels is already checked a few lines earlier.

libs/mve/image_io.cc
magcks: See L578.

PR #431, opened 2018-05-23
Fixed intrinsic

nitrotm:
Implementation for fixed intrinsic in makescene and sfmrecon.

- makescene has a new parameter effective when importing images (-i): "--fixed-intrinsic f,k1,k2,px,py,pa". For instance "--fixed-intrinsic 0.8,0.1,-0.1,0.5,0.5,1.0". This will simply fill up the view's meta.ini
- load/store "radial_distortion" parameter in view's meta.ini
- sfmrecon with "--fixed-intrinsics --intrinsics-from-views" will now use focal_length and radial_distortion
- partial support for principal-point parameter (only forwarded, not used yet)

To be discussed/improved:
- how to embed principal point parameter in computations: undistort_feature() and invalidate_large_error_tracks()?
- do we need to forward also pixel_aspect?

### DIFF
--- a/libs/mve/view.cc
+++ b/libs/mve/view.cc
@@ -139,6 +139,8 @@ View::load_view_from_mve_file  (std::string const& filename)
             && tokens.size() >= 2 && tokens.size() <= 7)
         {
             this->set_value("camera.focal_length", tokens[1]);
+            if (tokens.size() > 2)
+                this->set_value("camera.radial_distortion", tokens.concat(2, 2));
             if (tokens.size() > 4)
                 this->set_value("camera.pixel_aspect", tokens[4]);
             if (tokens.size() > 6)
@@ -382,6 +384,9 @@ View::set_camera (CameraInfo const& camera)
     /* Re-generate the "camera" section. */
     this->set_value("camera.focal_length",
         util::string::get_digits(camera.flen, 10));
+    this->set_value("camera.radial_distortion",
+        util::string::get_digits(camera.dist[0], 10) + " "
+        + util::string::get_digits(camera.dist[1], 10));
     this->set_value("camera.pixel_aspect",
         util::string::get_digits(camera.paspect, 10));
     this->set_value("camera.principal_point",
@@ -592,6 +597,7 @@ View::load_meta_data (std::string const& path)
 
     /* Get camera data from key/value pairs. */
     std::string cam_fl = this->get_value("camera.focal_length");
+    std::string cam_dist = this->get_value("camera.radial_distortion");
     std::string cam_pa = this->get_value("camera.pixel_aspect");
     std::string cam_pp = this->get_value("camera.principal_point");
     std::string cam_rot = this->get_value("camera.rotation");
@@ -600,6 +606,12 @@ View::load_meta_data (std::string const& path)
     this->meta_data.camera = CameraInfo();
     if (!cam_fl.empty())
         this->meta_data.camera.flen = util::string::convert<float>(cam_fl);
+    if (!cam_dist.empty())
+    {
+        std::stringstream ss(cam_dist);
+        ss >> this->meta_data.camera.dist[0];
+        ss >> this->meta_data.camera.dist[1];
+    }
     if (!cam_pa.empty())
         this->meta_data.camera.paspect = util::string::convert<float>(cam_pa);
     if (!cam_pp.empty())
@@ -692,7 +704,7 @@ View::populate_images_and_blobs (std::string const& path)
             proxy.name = name;
             this->blobs.push_back(proxy);
         }
-        else
+        else if (ext4 != ".ply") // silently ignore ply files
         {
             std::cerr << "View: Unrecognized extension "
                 << file.name << ", skipping." << std::endl;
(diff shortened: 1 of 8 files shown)


nitrotm:
I just realized that I have to adapt the braces code style. But I'll wait your review before going further.

libs/mve/view.cc
```diff
@@ -139,6 +139,8 @@ View::load_view_from_mve_file  (std::string const& filename)
             && tokens.size() >= 2 && tokens.size() <= 7)
         {
             this->set_value("camera.focal_length", tokens[1]);
+            if (tokens.size() > 2)
```
nitrotm: Should be > 3

libs/sfm/bundler_incremental.cc
```diff
@@ -330,6 +334,8 @@ Incremental::triangulate_new_tracks (int min_num_views)
                 continue;
             Viewport const& viewport = this->viewports->at(view_id);
             int const feature_id = track.features[j].feature_id;
+            // TODO: implement viewport principal point
```
nitrotm: Here needs some mathematical thinkering

simonfuhrmann:
A few questions before I go into the details of the PR:
- The argument to `makescene` is `--fixed-intrinsics`, but all it does is fill the `meta.ini` file. I'm wondering if we should rename this argument. Maybe something like `--init-intrinsics`.
- It would be a bit easier to review if you split some changes out into individual PRs, e.g., the changes to `makefile` are independent and useful even without the other changes.
- I wonder if specifying the principal point is useful. We don't support optimizing the PP in BA, and it's not trivial to do and unlikely to happen. WDYT?

nitrotm:
Okay, no problem.

If you agree, I suggest to open 2 PR then:

- 1 PR for `makescene` + `view.cc` (saving radial distortion in meta.ini)
- 1 PR for forwarding radial distortion + pp to ba

The pp is a 'nice-to-have' feature. I think most reasonably good cameras have a normalized pp around 0.5 (for instance, after calibration, mine is 0.50389 and 0.49676). So honestly I don't know if we can gain much by adding this extra parameter in ba.

But for the sake of consistency - the K matrix in the ba `Pose` contains px/py - I think it would be nice to use it in the two functions I mentioned (`undistort_feature` and `invalidate_large_error_tracks`). That means if the intrinsics are known and properly set in meta.ini (either by `--init-intrinsics` or by another mean), the camera pose optimisation (R,t) would take it into account. In the case the intrinsics are not given and fixed by user request, the current situation applies (f,k1,k2,R,t are optimized).

simonfuhrmann:
Adding PP to BA will introduce two additional degrees of freedom for the optimization, which does not only introduce more ambiguity for BA which can result in more unstable optimization, but also increases the complexity of the derivatives. I am unsure if using the PP in the undistort and invalidate functions, but not in the actual optimization, would introduce more harm than ignoring it completely. Maybe @flanggut has more insight...

And that ultimately begs the question if we should have it in the SfM code at all.

nitrotm:
I think it's fine to keep the PP fixed during BA. It doesn't need to be optimized and is by default at the image center (0.5, 0.5).

After looking at the code more in details, I am unsure where the principal point should apply. It seems the 2d feature points are already in normalized coordinates before triangulation, because the px,py of the K matrix are currently set to 0 (so effectively no translation from the image center).

Should the pp apply before that, in the feature detection stage then? Or maybe not at all? Again, I guess it's rare to have cameras where this parameter is not 0.5,0.5. Maybe cheaper hardware, such as cell phone cameras, might have this issue.

simonfuhrmann:
As you noticed, we normalize the coordinate system for SfM. 

Normalization happens here:
https://github.com/simonfuhrmann/mve/blob/master/libs/sfm/bundler_features.cc#L70

So it has to happen there, in the function call (but the function is not aware of any PP yet) or after the function call (some sort of correction step if the PP is known, and `--intrinsics-from-views` is set).

For now, let's get the `makescene` change in first. Can you create a separate PR for that?

simonfuhrmann:
Nevermind, just saw the `makescene` change.

nitrotm:
Ok, looking into `FeatureSet::normalize_feature_positions`, I see what seems a fixed PP setting currently:
```
        pos[0] = (pos[0] + 0.5f - fwidth / 2.0f) / fnorm;
        pos[1] = (pos[1] + 0.5f - fheight / 2.0f) / fnorm;
```
Would it be as simple as making this dynamic based on the pp parameter?

nitrotm:
Would it be correct to write (where px/py are normalized principal point factors):
```
        pos[0] = (pos[0] + 0.5f - fwidth * px) / fnorm;
        pos[1] = (pos[1] + 0.5f - fheight * py) / fnorm;
```
What is the 0.5f bias for in this formula?

simonfuhrmann:
The 0.5 is to convert from pixel coordinates in image space (i.e., integer numbers are pixel centers) to pixel coordinates in 3D space on the image plane (pixel centers have a .5). That's because the image plane starts at 0, ends at width/height, and the first pixel center is at 0.5/0.5.

I think your formula is correct.

nitrotm:
I've opened the second merge request #436

nitrotm:
These changes have been merged in PR #434 and #436.

simonfuhrmann:
Thanks for your contributions!

shikeqin123:
cool

PR #434, opened 2018-05-25
Allow to initialize camera intrinsics in makescene

nitrotm:
Implementation for fixed intrinsics in makescene.

- makescene has a new parameter effective when importing images (with -i): "--init-intrinsics f,k1,k2,ppx,ppy,pa". For instance "--init-intrinsics 0.8,0.1,-0.1,0.5,0.5,1.0". This will simply fill up the view's meta.ini
-  load/store "radial_distortion" parameter in view's meta.ini

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -899,9 +899,13 @@ import_images (AppSettings const& conf)
 
             tokenizer.split(conf.init_intrinsics, ',');
 
-            camera.flen = tokenizer.get_as<float>(0);
-            camera.dist[0] = tokenizer.get_as<float>(1);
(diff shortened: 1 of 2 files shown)


apps/makescene/makescene.cc
```diff
@@ -890,6 +891,27 @@ import_images (AppSettings const& conf)
...
 
+        /* Set initial intrinsics */
+        if (!conf.init_intrinsics.empty())
+        {
+            util::Tokenizer tokenizer;
+            mve::CameraInfo camera;
+
+            tokenizer.split(conf.init_intrinsics, ',');
+
+            camera.flen = tokenizer.get_as<float>(0);
```
simonfuhrmann: You need checks here if index 0, 1 and 2 is available.

PR #436, opened 2018-05-29
Use principal point in feature normalization and radial distortion in BA

nitrotm:
The second PR related to integrating fixed intrinsics in BA.

- Principal point is used in feature normalization (defaults to 0.5)
- Radial distortion is used in BA, either as a starting point or as a fixed parameter (defaults to 0.0)


simonfuhrmann:
Thanks

PR #441, opened 2018-08-08
Allow MVE to run on older hardware

pierotofy:
Hello there :hand:

Currently the default build settings will force MVE to be compiled using instruction sets that sometimes are not available to the CPU. Unless I've missed something, wouldn't specifying `march=native` take care of selecting the best available set of instructions for the user?

https://stackoverflow.com/questions/10686638/whats-the-differrence-among-cflgs-sse-options-of-msse-msse2-mssse3-msse4
https://gcc.gnu.org/onlinedocs/gcc/x86-Options.html

I thought I'd open a PR, but if I've overlooked something please feel free to close it.

Thanks!

### DIFF
--- a/Makefile.inc
+++ b/Makefile.inc
@@ -1,7 +1,7 @@
 # Default values.
 CXX ?= g++
 CXXWARNINGS ?= -Wall -Wextra -Wundef -pedantic
-CXXINTRINSICS ?= -msse2 -msse3 -msse4 -mpopcnt
+CXXINTRINSICS ?= -march=native
 CXXFEATURES ?= -funsafe-math-optimizations -fno-math-errno -std=c++11
 CXXFLAGS ?= ${CXXWARNINGS} ${CXXINTRINSICS} ${CXXFEATURES} -g -O3
 


simonfuhrmann:
Tricky... There are two potential issues I'm unsure about. 

- What if a certain SSE flag is not set, say we use a SSE4 instruction that is not set because of the native hardware, what will happen?
- What if MVE packages are compiled (read cross-compiled) for other machines. Then the `-march=native` is no good, but that may be a minor issue

Any thoughts?

pierotofy:
Good points, thanks for the reply!

I'm not familiar enough with the entire MVE codebase, but unless you are using `asm` directives I don't think point #1 would be an issue. The compiler generates opcodes based on the available architecture, so if SSE4 instructions are not available, they are replaced with less performant instructions.

Point #2 is spot on. If you plan to compile and distribute binaries, `native` won't cut it and typically you would compile a binary with an older instruction set (say, `-march=i686`). :+1:

simonfuhrmann:
I'm not aware of anyone bundling MVE so this is acceptable. Thank you for the contribution.

fragofer:
Hi @simonfuhrmann, I think that creating a tag/release with a working branch it would very helpful for external projects that are using your code.
Thanks!

simonfuhrmann:
Hi @fragofer, we don't have releases and just assume commits don't break the code. What is your concern?

PR #450, opened 2018-10-29
APPS: also build, clean and link apps/prebundle/prebundle

erictapen:
By building the projekt with `make`, `prebundle` is not built at the moment. Is this intended?


simonfuhrmann:
Looks like Googletest is broken :(

PR #473, opened 2019-04-06
Skip invalid views

pierotofy:
Hello :hand:

Sometimes when importing a scene via `makescene` an image might be corrupted. Instead of failing the process, it could be useful to allow the program to continue processing the rest of the scene while informing the user that a particular view could not be created.

This seems like it was the original intention per the comment to "handle exceptions while loading images".

Alternatively we could add a flag that specifies whether `makescene` should fail on import (the current behavior) or be more tolerant (the proposed change). Happy to make changes if needed / requested. :clinking_glasses:


simonfuhrmann:
Thanks!

Issue #476, opened 2019-05-05
Race condition and possible heap corruption


pierotofy:
Symptoms: sporadically `dmrecon` will crash with the following: https://gist.github.com/pierotofy/c49447e86a187e8ede50fb302cf5a47b

I've been able to get a core dump and the stack trace fails at https://github.com/simonfuhrmann/mve/blob/master/libs/dmrecon/patch_sampler.cc#L109 which is really odd because there's nothing that would suggest a problem around that line, and perhaps the presence of some odd memory corruption. 

```
(gdb) frame 13
#13 mvs::PatchSampler::fastColAndDeriv (this=<optimized out>, 
    v=<optimized out>, color=std::vector of length 0, capacity 0, 
    deriv=std::vector of length 0, capacity 0) at patch_sampler.cc:109
109	    std::vector<math::Vec2f> gradDir(nrSamples);
(gdb) info locals
refV = std::shared_ptr (count 3, weak 0) 0x0
nfp = <optimized out>
ratio = <optimized out>
mmLevel = <optimized out>
p1 = {static dim = <optimized out>, v = {698.763428, 431.70462, 106.666252}}
img = std::shared_ptr (count 2, weak 0) 0x0
gradDir = std::vector of length 0, capacity 0
(gdb) 
```

Here's the full core dump stack.

```
#0  0x00007fa2b9061428 in __GI_raise (sig=sig@entry=6)
    at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007fa2b906302a in __GI_abort () at abort.c:89
#2  0x00007fa2b90a37ea in __libc_message (do_abort=do_abort@entry=2, 
    fmt=fmt@entry=0x7fa2b91bced8 "*** Error in `%s': %s: 0x%s ***\n")
    at ../sysdeps/posix/libc_fatal.c:175
#3  0x00007fa2b90ae970 in malloc_printerr (ar_ptr=0x7fa2b93f0b20 <main_arena>, 
    ptr=0x1706c30, str=0x7fa2b91b9c92 "corrupted double-linked list", 
    action=<optimized out>) at malloc.c:5006
#4  _int_malloc (av=av@entry=0x7fa2b93f0b20 <main_arena>, 
    bytes=bytes@entry=200) at malloc.c:3727
#5  0x00007fa2b90b0184 in __GI___libc_malloc (bytes=200) at malloc.c:2913
#6  0x00007fa2b9de1e78 in operator new(unsigned long) ()
[...]

Issue #323, opened 2017-04-10
Segmentation Fault


goetzmoritz:
Hi, I'm running MVE on an ubuntu EC2 C4 8X Large with 100 images. Building went without failure.

When running dmrecon I ran into a segmentation fault. Here is what I got:

Received signal SIGSEGV (segmentation fault)
Obtained 15 stack frames: 0x4707b4 0x7fc918ebf4b0 0x40a20e 0x433070 0x4334ba 0x433560 0x42e4c7 0x427f33 0x40f596 0x413d35 0x40acd0 0x40bf97 0x7fc91969543e 0x7fc91925a6ba 0x7fc918f9082d
dmrecon/dmrecon[0x4707b4]
/lib/x86_64-linux-gnu/libc.so.6(+0x354b0)[0x7fc918ebf4b0]
dmrecon/dmrecon[0x40a20e]
dmrecon/dmrecon[0x433070]
dmrecon/dmrecon[0x4334ba]
dmrecon/dmrecon[0x433560]
dmrecon/dmrecon[0x42e4c7]
dmrecon/dmrecon[0x427f33]
dmrecon/dmrecon[0x40f596]
dmrecon/dmrecon[0x413d35]
dmrecon/dmrecon[0x40acd0]
dmrecon/dmrecon[0x40bf97]
/usr/lib/x86_64-linux-gnu/libgomp.so.1(+0xf43e)[0x7fc91969543e]
/lib/x86_64-linux-gnu/libpthread.so.0(+0x76ba)[0x7fc91925a6ba]
/lib/x86_64-linux-gnu/libc.so.6(clone+0x6d)[0x7fc918f9082d]
Segmentation fault
MVE Scene to Pointset (built on Apr 10 2017, 19:58:47)
Using depthmap "depth-L2" and color image "undist-L2"
Initializing scene with 100 views...
Initialized 100 views (max ID is 99), took 7ms.

Any suggestions?

simonfuhrmann:
Can you please provide a more complete `dmrecon` output also with the arguments you used?

goetzmoritz:
Unfortunatly not as this happend on EC2 and I actually just copy pasted what you see in my initial post.
I used "dmrecon -s2"

This is what I did:
I used 10 Images in a first test and run the pipeline until dmrecon -s2.
I checked the output which was fine and than uploaded another 100 images and run it again.
I deleted all the data beside the images.
Then I got the segmentation fault and the output was identical to the output of the 10 images.

[...]

PR #477, opened 2019-05-05
Added mutex lock to View::save_view() and View::cache_clean()

pierotofy:
This is a proposed fix for #476 and #323. The rationale for adding a mutex on save_view and cache_clean is that `cache_clean` seems to be the function responsible for deallocating the pointers too early in certain race condition instances, while the corrupted pointers show up in `save_image_intern`, which is only ever called by `save_view`.

I don't know if this is the best way to fix it, but seems to work. The verification has been done by running the code modified for triggering the fault (https://github.com/pierotofy/mve/tree/segfaultpoc) (`./test1.sh` contains code to re-run `dmrecon` until it fails). Before this patch the code would fail before ~5-10 iterations. After this patch the code seems to have stopped seg faulting.

Feedback is welcome. :clinking_glasses:

### DIFF
--- a/libs/mve/view.cc
+++ b/libs/mve/view.cc
@@ -225,6 +225,8 @@ View::save_view_as (std::string const& user_path)
 int
 View::save_view (void)
 {
+    std::lock_guard<std::mutex> lock(this->sv_cc_mutex);
+
     if (this->path.empty())
         throw std::runtime_error("View not initialized");
 
@@ -305,6 +307,8 @@ View::is_dirty (void) const
 int
 View::cache_cleanup (void)
 {
+    std::lock_guard<std::mutex> lock(this->sv_cc_mutex);
+        
     int released = 0;
     for (std::size_t i = 0; i < this->images.size(); ++i)
     {
--- a/libs/mve/view.h
+++ b/libs/mve/view.h
@@ -47,6 +47,7 @@
 #include <memory>
 #include <string>
 #include <vector>
+#include <mutex>
 
 #include "mve/defines.h"
 #include "mve/camera.h"
@@ -300,6 +301,7 @@ private:
     void load_blob_intern (BlobProxy* proxy, bool init_only);
     void save_blob_intern (BlobProxy* proxy);
 
+    std::mutex sv_cc_mutex; // protect save_view and cache_clean
 protected:
     typedef std::vector<std::string> FilenameList;
 


simonfuhrmann:
First of all, thank you so much for investigating this. I was reading #476 and this bug seems to be a tricky one. Multi-threading issues are a beast. From your debug print stagements:

````
F 0x7f5548003860
ipc_cleanup 1:end
...
cache_cleanup: start
G 0
````

... it seems obvious that there is something going on between saving views and cache_cleanup, if both of the `F` and `G` lines are printed from the same view (can you confirm it's definitely printed from the same view?).

Regarding the fix, I am happy to pull this if we are desperate enough to just get this fixed. However let me note a few things. 1. The mve::View class is currently not thread safe, and is not supposed to be thread safe. Even after this fix, there are probably a lot of threading issues, and the solution generally is to not call `cache_cleanup()` which `save_view()` is also running. I'm wondering how this can ever happen. Do you have some insight in this?

Just by looking at the code, `save_view()` is only ever called in the `dmrecon.cc` app. The other call, `cache_cleanup()`, is only called in `image_pyramid.cc` after an image is loaded from a neighboring view to create the image pyramid. And that does sound like a real issue, because that neighboring view could be a view that's currently saved.

However. `cache_cleanup()` should not touch embeddings that are dirty. So my question is, does this suggestion fix the problem as well: Move line 877 in `view.cc` (`proxy->is_dirty = false;`) a little below, say at the end of the function. Then, things should not get cleaned up before the save is complete.

Again, thank you for investigating this. This is really good Sherlock Holmes work here.

**Comment by pierotofy** [comment]
Thanks for the reply! Tricky indeed! I think it just made my top 10 list of most difficult bugs ever.

I've added unique IDs to the views, and the output showed:

```
E:1521 0x7fd517c77fb0
ipc_cleanup 11:start
cache_cleanup: start
F:1521 0x7fd517c77fb0
cache_cleanup: end
ipc_cleanup 11:end
ipc_cleanup 0:start
cache_cleanup: start
cache_cleanup: end
ipc_cleanup 0:end
ipc_cleanup 1:start
cache_cleanup: start
G:1521 0
```

So F and G are from the same view.

Moving the expression (and 5 other similar ones for good measure) also seems to prevents the seg fault. The logic makes sense, I had a feeling that there would be a better way, but I'm still learning the MVE codebase.

I've opened #478 for review. If that looks OK, feel free to close this one. 

Thank you so much for the suggestion!

simonfuhrmann:
Closed by #478

PR #478, opened 2019-05-06
Fix race condition

pierotofy:
Alternate, better fix than the one proposed in #477.


simonfuhrmann:
Perfect, thank you so much. I would award you an "achievement unlocked" if I could. :)

pierotofy:
Thanks for reviewing the changes. Happy to contribute :+1:

PR #483, opened 2019-08-23
Add Libexecinfo flag for alpine + create Dockerfile

npcasler:
Alpine doesn't include the backtrace functionality of glibc, which is
used in libs/util/system.cc. Libexecinfo provides the missing function
symbols for Alpine and FreeBSD distros. Adding a check in Makefile.inc
for /etc/alpine-release will determine if the library needs to be
linked.

Provide a Dockerfile to validate the build on Alpine. The Dockerfile
provides a minimal set of dependencies needed for compilation.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,17 @@
+FROM alpine:3.10
+
+COPY . /mve
+
+RUN apk add --no-cache \
+    make \
+    g++ \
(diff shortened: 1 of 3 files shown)


npcasler:
Roger that, new PR created. Closing this PR.

PR #488, opened 2019-08-29
Check for GLIBC rather than GNUC for compatibility

npcasler:
The current compatibility checks in the libs/util/system.cc
check for the __GNUC__ flag when deciding whether to include
the execinfo.h header and backtrace functionalityt. The
__GNUC__ flag returns true for distros that do not inlude glibc
which will cause compilation issues for musl libc.

The check now looks for __GLIBC__ and will properly exclude the
functionality from distros using a different libc implementation.

Also a Dockerfile is included which can be used to validate builds
on musl based distros such as Alpine.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,16 @@
+ARG ALPINE_VERSION=3.10
+FROM alpine:${ALPINE_VERSION}
+
+COPY . /mve
+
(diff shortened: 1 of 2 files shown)


simonfuhrmann:
This looks good, thanks! I hope this works for you.

PR #493, opened 2019-10-23
Add support for TIFF files in load_file_headers

pierotofy:
Hello :hand:

This PR adds support for TIFF files in the `load_file_headers`, which is currently missing.

Hope this is useful for others. :+1:

Please let me know if changes are needed.

### DIFF
--- a/libs/mve/image_io.cc
+++ b/libs/mve/image_io.cc
@@ -643,8 +643,8 @@ load_tiff_file_headers (std::string const& filename)
     TIFFSetErrorHandler(tiff_error_handler);
 
     TIFF* tif = TIFFOpen(filename.c_str(), "r");
-    if (!tif)
-        throw util::Exception("TIFF file format not recognized");
+    if (tif == nullptr)
+        throw util::FileException(filename, "TIFF file format not recognized");
 
     try
     {
@@ -661,8 +661,8 @@ load_tiff_file_headers (std::string const& filename)
         headers.channels = channels;
         headers.type = IMAGE_TYPE_UNKNOWN;
 
-        if (bits == 8){
-            switch(sampleFormat){
+        if (bits == 8) {
+            switch(sampleFormat) {
                 case SAMPLEFORMAT_UINT:
                     headers.type = IMAGE_TYPE_UINT8;
                     break;
@@ -675,8 +675,8 @@ load_tiff_file_headers (std::string const& filename)
                 default:
                     break;
             }
-        }else if (bits == 16){
-            switch(sampleFormat){
+        } else if (bits == 16) {
+            switch(sampleFormat) {
                 case SAMPLEFORMAT_UINT:
                     headers.type = IMAGE_TYPE_UINT16;
                     break;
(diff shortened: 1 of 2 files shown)


libs/mve/image_io.cc
```diff
@@ -628,6 +635,104 @@ tiff_error_handler (char const* /*module*/, char const* fmt, va_list ap)
...
 
+ImageHeaders
+load_tiff_file_headers (std::string const& filename)
+{
+    ImageHeaders headers;
+    TIFFSetWarningHandler(nullptr);
+    TIFFSetErrorHandler(tiff_error_handler);
+
+    TIFF* tif = TIFFOpen(filename.c_str(), "r");
+    if (!tif)
```
simonfuhrmann: `if (tif == nullptr)`

libs/mve/image_io.cc
```diff
@@ -628,6 +635,104 @@ tiff_error_handler (char const* /*module*/, char const* fmt, va_list ap)
...
+        TIFFGetField(tif, TIFFTAG_SAMPLESPERPIXEL, &channels);
+        TIFFGetField(tif, TIFFTAG_BITSPERSAMPLE, &bits);
+        TIFFGetField(tif, TIFFTAG_SAMPLEFORMAT, &sampleFormat);
+
+        headers.width = width;
+        headers.height = height;
+        headers.channels = channels;
+        headers.type = IMAGE_TYPE_UNKNOWN;
+
+        if (bits == 8){
```
simonfuhrmann: space before `{` here and below

libs/mve/image_io.cc
```diff
@@ -628,6 +635,104 @@ tiff_error_handler (char const* /*module*/, char const* fmt, va_list ap)
...
+                case SAMPLEFORMAT_INT:
+                    headers.type = IMAGE_TYPE_SINT8;
+                    break;
+                case SAMPLEFORMAT_IEEEFP:
+                    headers.type = IMAGE_TYPE_FLOAT;
+                    break;
+                default:
+                    break;
+            }
+        }else if (bits == 16){
```
simonfuhrmann: space after `}` here and below

libs/mve/image_io.cc
```diff
@@ -628,6 +635,104 @@ tiff_error_handler (char const* /*module*/, char const* fmt, va_list ap)
...
+ImageHeaders
+load_tiff_file_headers (std::string const& filename)
+{
+    ImageHeaders headers;
+    TIFFSetWarningHandler(nullptr);
+    TIFFSetErrorHandler(tiff_error_handler);
+
+    TIFF* tif = TIFFOpen(filename.c_str(), "r");
+    if (!tif)
+        throw util::Exception("TIFF file format not recognized");
```
simonfuhrmann: Would that be a `util::FileException`?

libs/mve/image_io.cc
pierotofy: Better, yes. Thank you.

libs/mve/image_io.cc
pierotofy: Good catch, yes it would be better instead of the generic one. :+1:

pierotofy:
Much appreciate the feedback!

PR #508, opened 2020-06-08
fix not supported file system cases

DaniilSNikulin:
from documentation for function `readdir`:
```
Currently, only some filesystems (among them: Btrfs, ext2, ext3,
and ext4) have full support for returning the file type in d_type.
All applications must properly handle a return DT_UNKNOWN
```
To avoid problems with not supported file system, `is_dir` flag additionally  computed via function `stat`, which don't have same constraints.

In my case, `DT_UNKNOWN` was given from `readdir`, if sub-folder is mounted via `sshfs`.


simonfuhrmann:
Thanks!

PR #509, opened 2020-07-29
Support for importing Colmap SfM models

SBCV:
This PR allows to import Colmap's SfM model formats (`.bin` as well as `.txt` files). 

Let us assume the following folder structure:
```
<Colmap workspace> 
    <images>
    <sparse model>
        cameras.txt / cameras.bin
        images.txt / images.bin 
        points3D.txt / points3D.bin
```

The importer looks on the same level than the `<sparse model>` for the  `<image folder>` with the corresponding images.
Also, the importer searches first for the `.txt` files in the `<sparse model>` folder. 

In order to validate the imported results, I performed a reconstruction with Colmap and exported the results as `.NVM` file as well as Colmap's standard model folders (.txt and .bin). Then, I compared the values in the `Bundle::Ptr` returned by `load_nvm_bundle` and `load_colmap_bundle`. I did not add the testing code, since I did not find a suitable place in the MVE library. 

If you have any questions or comments, please let me know.

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -337,15 +337,25 @@ make_image_name (int id)
 /* ---------------------------------------------------------------- */
 
 void
-import_bundle_nvm (AppSettings const& conf)
+import_bundle_nvm_or_colmap (AppSettings const& conf, bool load_nvm = true)
 {
-    std::vector<mve::NVMCameraInfo> nvm_cams;
-    mve::Bundle::Ptr bundle = mve::load_nvm_bundle(conf.input_path, &nvm_cams);
+    std::vector<mve::AdditionalCameraInfo> cam_infos;
+    mve::Bundle::Ptr bundle;
+    if (load_nvm)
+        bundle = mve::load_nvm_bundle(conf.input_path, &cam_infos);
+    else
+        bundle = mve::load_colmap_bundle(conf.input_path, &cam_infos);
+
     mve::Bundle::Cameras& cameras = bundle->get_cameras();
 
-    if (nvm_cams.size() != cameras.size())
+    if (cam_infos.size() != cameras.size())
     {
-        std::cerr << "Error: NVM info inconsistent with bundle!" << std::endl;
+        if (load_nvm)
+            std::cerr << "Error: NVM info inconsistent with bundle!" 
+                << std::endl;
+        else
+            std::cerr << "Error: Colmap info inconsistent with bundle!" 
+                << std::endl;
         return;
     }
 
@@ -360,7 +370,7 @@ import_bundle_nvm (AppSettings const& conf)
     for (std::size_t i = 0; i < cameras.size(); ++i)
     {
         mve::CameraInfo& mve_cam = cameras[i];
-        mve::NVMCameraInfo const& nvm_cam = nvm_cams[i];
+        mve::AdditionalCameraInfo const& cam_info = cam_infos[i];
         std::string fname = "view_" + util::string::get_filled(i, 4) + ".mve";
 
         mve::View::Ptr view = mve::View::create();
@@ -369,10 +379,10 @@ import_bundle_nvm (AppSettings const& conf)
 
         /* Load original image. */
         std::string exif;
-        mve::ImageBase::Ptr image = load_any_image(nvm_cam.filename, &exif);
+        mve::ImageBase::Ptr image = load_any_image(cam_info.filename, &exif);
         if (image == nullptr)
         {
-            std::cout << "Error loading: " << nvm_cam.filename
+            std::cout << "Error loading: " << cam_info.filename
                 << " (skipping " << fname << ")" << std::endl;
             continue;
         }
@@ -401,8 +411,8 @@ import_bundle_nvm (AppSettings const& conf)
         /* Add original image. */
         if (conf.import_orig)
         {
-            if (has_jpeg_extension(nvm_cam.filename))
-                view->set_image_ref(nvm_cam.filename, "original");
+            if (has_jpeg_extension(cam_info.filename))
+                view->set_image_ref(cam_info.filename, "original");
             else
                 view->set_image(image, "original");
         }
@@ -414,7 +424,9 @@ import_bundle_nvm (AppSettings const& conf)
         mve_cam.flen = mve_cam.flen / static_cast<float>(maxdim);
 
         mve::ByteImage::Ptr undist = mve::image::image_undistort_vsfm<uint8_t>
-            (std::dynamic_pointer_cast<mve::ByteImage const>(image), mve_cam.flen, nvm_cam.radial_distortion);
+            (std::dynamic_pointer_cast<mve::ByteImage const>(image), 
+            mve_cam.flen, 
+            cam_info.radial_distortion);
         undist = limit_image_size<uint8_t>(undist, conf.max_pixels);
         view->set_image(undist, "undistorted");
         view->set_camera(mve_cam);
@@ -785,6 +797,36 @@ is_noah_bundler_format (AppSettings const& conf)
     return util::fs::file_exists(bundle_fname.c_str());
 }
 
+bool
+is_colmap_sfm_bundle_format (AppSettings const& conf)
+{
+    std::string cameras_txt_filename = util::fs::join_path(conf.input_path, 
+        "cameras.txt");
+    std::string cameras_bin_filename = util::fs::join_path(conf.input_path, 
+        "cameras.bin");
+    if (!util::fs::file_exists(cameras_txt_filename.c_str()) && 
+        !util::fs::file_exists(cameras_bin_filename.c_str()))
+        return false;
+
+    std::string images_txt_filename = util::fs::join_path(conf.input_path, 
+        "images.txt");
+    std::string images_bin_filename = util::fs::join_path(conf.input_path, 
+        "images.bin");
+    if (!util::fs::file_exists(images_txt_filename.c_str()) && 
+        !util::fs::file_exists(images_bin_filename.c_str()))
+        return false;
+
+    std::string points_3D_txt_filename = util::fs::join_path(conf.input_path, 
+        "points3D.txt");
+    std::string points_3D_bin_filename = util::fs::join_path(conf.input_path, 
+        "points3D.bin");
+    if (!util::fs::file_exists(points_3D_txt_filename.c_str()) && 
+        !util::fs::file_exists(points_3D_bin_filename.c_str()))
+        return false;
+
+    return true;
+}
+
 /* ---------------------------------------------------------------- */
 
 void
@@ -797,7 +839,7 @@ import_bundle (AppSettings const& conf)
     if (is_visual_sfm_bundle_format(conf))
     {
         std::cout << "Info: Detected VisualSFM bundle format." << std::endl;
-        import_bundle_nvm(conf);
+        import_bundle_nvm_or_colmap(conf, true);
         return;
     }
 
@@ -815,6 +857,18 @@ import_bundle (AppSettings const& conf)
     {
         std::cout << "Info: Detected Noah bundler format." << std::endl;
         import_bundle_noah_ps(conf);
+        return;
+    }
+
+    /**
+     * Try to detect Colmap bundle format.
+     * In this case the input folder contains files with extension ".txt" or 
+     * ".bin".
+     */
+    if (is_colmap_sfm_bundle_format(conf))
+    {
+        std::cout << "Info: Detected colmap bundler format." << std::endl;
+        import_bundle_nvm_or_colmap(conf, false);
         return;
     }
 
--- a/libs/mve/bundle_io.cc
+++ b/libs/mve/bundle_io.cc
@@ -12,7 +12,10 @@
 #include <iostream>
 #include <string>
 #include <fstream>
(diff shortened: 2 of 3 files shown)


libs/mve/endian.h
```diff
@@ -0,0 +1,167 @@
+// Copyright (c) 2018, ETH Zurich and UNC Chapel Hill.
```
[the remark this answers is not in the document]
SBCV: > The first thing that comes to mind is that we already have endian code in lib/util/system.h

Ah nice, I was not aware of those functions - I'll have a look on Sunday / Monday and come back to you.

SBCV:
Once #510 is merged, I'll update this PR accordingly (i.e. rebasing this on the master branch)

SBCV:
Side note, I've used the addon mentioned in #512 to import a Colmap reconstruction and the corresponding MVE workspace using the functionality provided by this PR. The camera poses and points aligned perfectly.
Because of those results, I guess that the import of the extrinsic camera parameters should work as expected. 
For comparison purposes, I've shifted one reconstruction a bit in the image below. 
colmap_mve

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+    // https://github.com/colmap/colmap/blob/dev/src/base/camera_models.h
+    // https://github.com/simonfuhrmann/mve/blob/master/libs/mve/camera.cc
+    if (model == "SIMPLE_PINHOLE")
+    {
+        // Simple pinhole: f, cx, cy
+        camera_info.flen = params[0];
+        camera_info.ppoint[0] = params[1] / width;
+        camera_info.ppoint[1] = params[2] / height;
+    }
+    else if (model == "PINHOLE")
```
SBCV, line 545: However, I'm not 100% sure that the intrinsic parameter conversion is correct. In particular, I'm wondering if I correctly understood  the meaning of `paspect` ratio.

simonfuhrmann: `paspect` is the aspect ratio of the pixels, which should be 1.0 for almost all cameras. It will be >1.0 if pixels are wider than tall.

simonfuhrmann: If the focal length is different and x and y direction, that could only mean that pixels have different aspect ratio? If that's the case, your code looks reasonable.

SBCV:
Rebased and Squashed the commits

SBCV:
Another side note: I figured out that colmap exports nvm files with feature observation positions within `[0, width] x [0, height]`. However, nvm files created with VisualSfM contain features that are centered around the principal point, e.g. they lie within `[-width/2, width/2] x [-height/2, height/2]`. 
What is the correct range that MVE expects? Does MVE provide an option to visualize feature observations?

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+    }
+    in_cameras.close();
+}
+
+void
+load_colmap_images_txt(std::string const& images_filename,
+    std::string const& image_path,
+    std::map<uint32_t, CameraInfo>& camera_colmap_id_to_model,
+    Bundle::Ptr& bundle,
+    std::map< int, std::vector<Bundle::Feature2D> >& view_id_to_features_2d,
```
simonfuhrmann, line 605: Pass this as a pointer, to indicate at the call site that this gets modified.

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+        colmap_cams_info.push_back(colmap_cam_info);
+    }    
+    if (camera_info != nullptr)
+        std::swap(*camera_info, colmap_cams_info);
+    in_images.close();
+}
+
+void
+load_colmap_points_3D_txt(std::string const& points3D_filename,
+    std::map< int, std::vector<Bundle::Feature2D> >& view_id_to_features_2d,
```
simonfuhrmann, line 678: same here

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+            int feature_2d_idx;      // starts at 0
+            point_3d_line_ss >> view_colmap_id >> feature_2d_idx;
+            int view_id = view_colmap_id - 1;
+            // https://github.com/colmap/colmap/blob/dev/src/base/reconstruction.cc#L927
+            // Make sure that each point only has a single observation per image,
+            // since MVE does not support multiple observations.
+            if (std::find(view_ids.begin(), view_ids.end(), view_id) != view_ids.end())
+                continue;
+            view_ids.push_back(view_id);
+            Bundle::Feature2D corresponding_feature = view_id_to_features_2d.at(view_id).at(feature_2d_idx);
```
simonfuhrmann, line 726: Line break to 80 chars

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+        CameraInfo camera_info;
+        create_camera_info_from_params(
+            camera_info, model_name, params, width, height);
+        camera_colmap_id_to_info[camera_colmap_id] = camera_info;
+    }
+    in_cameras.close();
+}
+
+void
+load_colmap_images_bin(std::string const& images_filename,
```
simonfuhrmann, line 794: I assume you have looked into this, but this can't be merged with the txt version to avoid code duplication?

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+        rot = get_rot_from_quaternion(quat);
+        std::copy(rot.begin(), rot.end(), bundle_cam.rot);
+        std::copy(trans.begin(), trans.end(), bundle_cam.trans);
+
+        colmap_cam_info.filename = image_name;
+        colmap_cam_info.radial_distortion = model.dist[0];
+        if (!util::fs::is_absolute(colmap_cam_info.filename))
+            colmap_cam_info.filename = util::fs::join_path(image_path, colmap_cam_info.filename);
+
+        const std::size_t num_points_2D = util::system::read_binary_little_endian<uint64_t>(&in_images);
```
simonfuhrmann, line 858: Line breaks here and below

libs/mve/bundle_io.cc
```diff
@@ -480,4 +483,537 @@ save_photosynther_bundle (Bundle::ConstPtr bundle, std::string const& filename)
...
+    std::string cameras_txt_filename = util::fs::join_path(foldername, "cameras.txt");
+    std::string cameras_bin_filename = util::fs::join_path(foldername, "cameras.bin");
+
+    std::string images_txt_filename = util::fs::join_path(foldername, "images.txt");
+    std::string images_bin_filename = util::fs::join_path(foldername, "images.bin");
+
+    std::string points_3D_txt_filename = util::fs::join_path(foldername, "points3D.txt");
+    std::string points_3D_bin_filename = util::fs::join_path(foldername, "points3D.bin");
+
+    std::cout << "Colmap: Loading folder..." << std::endl;
```
simonfuhrmann, line 970: folder -> directory

simonfuhrmann:
> What is the correct range that MVE expects? Does MVE provide an option to visualize feature observations?

I believe in the bundle MVE expects [0,width]x[0,height]. In SfM code it's a little different and centered around 0 to apply radial distortion.

apps/makescene/makescene.cc
```diff
@@ -414,7 +422,7 @@ import_bundle_nvm (AppSettings const& conf)
         mve_cam.flen = mve_cam.flen / static_cast<float>(maxdim);
 
         mve::ByteImage::Ptr undist = mve::image::image_undistort_vsfm<uint8_t>
-            (std::dynamic_pointer_cast<mve::ByteImage const>(image), mve_cam.flen, nvm_cam.radial_distortion);
+            (std::dynamic_pointer_cast<mve::ByteImage const>(image), mve_cam.flen, cam_info.radial_distortion);
```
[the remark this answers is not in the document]
SBCV: Just for clarification: There must be no lines with more than 80 characters? (I'm asking this because the file contains lines written by other authors that are longer than 80 characters). No problem to make these adjustments, I just wanted to be sure.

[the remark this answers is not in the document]
SBCV: If so, I guess I touch only lines that are already modified by this PR?

libs/mve/bundle_io.cc
SBCV: According to p. 156 / 157 in "Multiple View Geometry" (Hartley and Zisserman) 
	`fx = f mx` and 
	`fy = f my`, 
where `mx` and `my` are "the number of pixels per unit distance in image coordinates [...] in x and y directions". 
Let us assume that the pixels are wider than tall (i.e. `paspect > 1`).
If pixels are wider than tall, then `my > mx` and `fy > fx`. 
But then `camera_info.paspect = fx / fy` would be `< 1`. 
Thus, I think we have to replace this computation with `camera_info.paspect = fy / fx`, which is `> 1` 
Does that make sense?

libs/mve/bundle_io.cc
SBCV: `view_id_to_features_2d` is not getting modified within `load_colmap_points_3D_txt()`- only in `load_colmap_images_txt()`. Or am I missing something here?

libs/mve/bundle_io.cc
SBCV: I think that the merged result would be inconvenient to write and understand. 
However, in the latest commit of today I've wrapped most parts of the redundant code into specific functions that are now re-used to avoid code duplication. What do you think?

libs/mve/bundle_io.cc
simonfuhrmann: What you are writing makes sense, the math is sound. But I am getting a little confused now...
Imagine a camera with square pixels, and `fx = fy`. If we now start making pixels less tall, i.e., shrink them vertically, `fy` will get smaller, but `paspect` should get larger, right? Hence `fy / fx` will produce the wrong result? 

As I said, I'm a little confused at this point, I might be missing something.

simonfuhrmann:
Would you mind squashing and force commiting the PR? It'll get a little easier to review.
I'm not entirely happy how Github's code review works, or I just don't know how to work with it very well... is there a way to review all effective changes in one diff?

SBCV:
No problem. I'm afraid, I'm not an github review expert either :o 
Is it clearer now?

libs/mve/bundle_io.cc
SBCV: I understand it this way: 
Assuming we have squared pixel (fx = fy) and the size of these squared pixels define the units of some coordinate system (i.e. `my = 1`). In the following `fy'` and `my'` denote the values of the shrinked pixels. If we shrink the pixels vertically, then more pixels actually fit in a unit distance (`my' > 1`). Remember: `mx` and `my` are "the NUMBER of pixels per unit distance in image coordinates [...] in x and y directions". Since `my' > 1`, it follows that `fy' > fy` (i.e. `fy` becomes bigger).
Or in other words, we need more of the shrinked pixels to express the original focal length `f`` (`f = fy / my`).

But I could be wrong here. Since you are the maintainer, you should decide how we handle this.

Side note: I've wrote this importer to compute textured reconstructions of satellite images (the corresponding SfM reconstruction contains camera models with different focal length values fx and fy). With these reconstructions, I hope I am be able to empirically verify this decision :D

simonfuhrmann:
I'm merging this for now. I'm still not convinced with the `paspect` thing, so it would be great it you could empirically validate this (see which one of the two versions produce better results?) :)

simonfuhrmann:
Thanks for your contribution!

SBCV:
Thx for merging!

SBCV:
I've done some experiments and could verify that  `paspect = fy / fx` is correct. 

Here a result with  `paspect = fy / fx`
image

And here a result with  `paspect = fx / fy`
image

Btw: 
Both branches of the if-else case in [camera.cc](https://github.com/simonfuhrmann/mve/blob/master/libs/mve/camera.cc#L130) use also (implicit) the relation: `paspect = fy / fx`

simonfuhrmann:
Thank you for confirming it works. This is a nice reconstruction. What was the software process for reconstructing this? How was the imagery taken?

simonfuhrmann:
Also, have you tried using UMVE for mesh inspection? It doesn't have a ton of features, but it's faster than Meshlab.

SBCV:
> Thank you for confirming it works. This is a nice reconstruction. What was the software process for reconstructing this? How was the imagery taken?

The sfm part has been performed with this [colmap fork](https://github.com/Kai-46/VisSatSatelliteStereo). Other pre- and post processing steps have been performed with some custom code. The imagery is part of https://spacenetchallenge.github.io/datasets/mvs_summary.html

> Also, have you tried using UMVE for mesh inspection? It doesn't have a ton of features, but it's faster than Meshlab.

Oh, I wasnt aware of that. Thanks for pointing out.

PR #510, opened 2020-08-02
Added reading little endian functions

SBCV:
Adds two utility functions to read little endian binary streams. 
This PR serves as preparation for #509.

### DIFF
--- a/libs/util/system.h
+++ b/libs/util/system.h
@@ -14,6 +14,7 @@
 #include <cstdlib>
 #include <thread>
 #include <chrono>
+#include <vector>
 
 #include "util/defines.h"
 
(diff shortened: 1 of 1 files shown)


libs/util/system.h
```diff
@@ -238,6 +239,20 @@ betoh (T const& x)
 #   error "Couldn't determine host endianess!"
 #endif
 
+template <typename T>
```
simonfuhrmann, line 242: Please include istream.
Please format the functions according to the style of the rest of the functions, i.e., indentation and bracket position.
Please also add the function signature above (below the betoh signature). These files are generally organized to have signatures on top, and implementation here.

libs/util/system.h
```diff
@@ -238,6 +239,20 @@ betoh (T const& x)
...
+T read_binary_little_endian(std::istream* stream) {
+  T data_little_endian;
+  stream->read(reinterpret_cast<char*>(&data_little_endian), sizeof(T));
+  return letoh(data_little_endian);
+}
+
+template <typename T>
```
simonfuhrmann, line 249: This is an odd function that modifies the input vector yet expects it to be initialized to a certain size. This also doesn't seem very versatile, and perhaps shouldn't be in system.h. Can you move that function to the place where you actually use it (to your other PR)?

libs/util/system.h
SBCV: Ouch, how could I miss that.

PR #514, opened 2020-08-20
Extension of Colmap Importer

SBCV:
Improvements regarding #509

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -25,6 +25,7 @@
 #include <algorithm>
 #include <atomic>
 #include <unordered_map>
+#include <cassert>
 
 #include "math/matrix.h"
 #include "math/matrix_tools.h"
@@ -72,6 +73,7 @@ struct AppSettings
     bool append_images = false;
     int max_pixels = std::numeric_limits<int>::max();
     std::string init_intrinsics;
+    int scale = -1;
 
     /* Computed values. */
     std::string bundle_path;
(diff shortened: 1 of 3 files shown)


SBCV:
This PR improves #509 in the following ways:

1.) It adds an option to load Colmap depth maps into MVE. There is a scale option, which allows to downsample the depth maps to a specific scale. So that it can be used in a similar fashion as MVE's `dm_recon` functionality. This allows to combine Colmap's MVS step with MVE's Floating Scale Surface Reconstruction algorithm. 

2.) As of #509, images corresponding to camera models of type `Simple radial` are undistorted with VisualSfM specific undistortion algorithm (see [here](https://github.com/simonfuhrmann/mve/blob/master/apps/makescene/makescene.cc#L426)). This leads to incorrect results because of different semantics of radial distortion coefficients of Colmap and VisualSfM. 
Thus, the importer verifies that the Colmap reconstruction contains only camera models without radial distortion coefficients. To provide suitable results, one can execute the first/undistortion step in Colmap's MVS pipeline. This undistorts the images and the corresponding camera models have no radial distortion coefficients.

apps/makescene/makescene.cc
```diff
@@ -423,12 +426,55 @@ import_bundle_nvm_or_colmap (AppSettings const& conf, bool load_nvm = true)
...
+                (std::dynamic_pointer_cast<mve::ByteImage const>(image),
+                mve_cam.flen,
+                cam_info.radial_distortion);
+        }
+        else
+        {
+            // For Colmap reconstructions, MVE expects camera models without
+            // radial distortion. Potentially, obtained in Colmap's undistortion
+            // step.
+            assert(cam_info.radial_distortion == 0.0);
```
simonfuhrmann, line 442: I'm not sure if asserting provides appropriate communication to the user. Should this instead be an if, a message, and an std::exit(1) or something?

apps/makescene/makescene.cc
```diff
@@ -800,30 +846,32 @@ is_noah_bundler_format (AppSettings const& conf)
 bool
 is_colmap_sfm_bundle_format (AppSettings const& conf)
 {
-    std::string cameras_txt_filename = util::fs::join_path(conf.input_path, 
-        "cameras.txt");
-    std::string cameras_bin_filename = util::fs::join_path(conf.input_path, 
-        "cameras.bin");
+    using util::fs::join_path;
+    std::string workspace_path = conf.input_path;
+    std::string model_path(join_path(workspace_path, "sparse"));
```
simonfuhrmann, line 851: Prefer assignment instead of constructor. It's equivalent, but semantically clearer. It's unheard of that people initialize their primitives with the value constructor `int x(10)` for the same reason.

apps/makescene/makescene.cc
SBCV: This assertion is really considered as assertion. It should be always true, since other cases are detected earlier in [ libs/mve/bundle_io.cc at line 565](https://github.com/simonfuhrmann/mve/pull/514/commits/eec565e886899815eb88e294e8e19374b77f463b#diff-b48aa0fdef49e807586f79bef370769dL565). 
     
```
        std::string msg = "Unsupported camera model with radial distortion "
            "detected! If possible, re-run the SfM reconstruction with the "
            "SIMPLE_PINHOLE or the PINHOLE camera model (recommended). " 
            "Otherwise, use the undistortion step in Colmap to obtain "
            "undistorted images and corresponding camera models without radial "
            "distortion.";

        throw util::Exception(msg.c_str());
```

Is throwing an exception there okay? Or should it be replaced with `std::exit(1)`?
Side note: Other functions in `bundle_io.cc` use exceptions for similar cases (and not `std::exit(1)`)

simonfuhrmann: If this is really just an assertion, this is fine then.

simonfuhrmann: On another note, exceptions are a thing for libraries, and we don't want to call std::exit() in a library. This is a CLI, and hence responsible for catching all exceptions and making sure a nice message is printed to the user.

SBCV:
Added requested changes and squashed the commit.
Somehow Github still reports that there is an open change request - but I do not see where

PR #518, opened 2020-09-07
Fix importer intrinsics

SBCV:
Fix Colmap importer intrinsic conversion and remove redundant depth map conversion code

### DIFF
--- a/apps/makescene/makescene.cc
+++ b/apps/makescene/makescene.cc
@@ -422,8 +422,12 @@ import_bundle_nvm_or_colmap (AppSettings const& conf, bool load_nvm = true)
         add_exif_to_view(view, exif);
 
         /* Normalize focal length, add undistorted image. */
-        int const maxdim = std::max(image->width(), image->height());
-        mve_cam.flen = mve_cam.flen / static_cast<float>(maxdim);
+        if (load_nvm)
+        {
+            // In the case of Colmap, the focal length is already normalized
+            int const maxdim = std::max(image->width(), image->height());
+            mve_cam.flen = mve_cam.flen / static_cast<float>(maxdim);
+        }
 
         mve::ByteImage::Ptr undist;
         if (load_nvm)
--- a/libs/mve/bundle_io.cc
+++ b/libs/mve/bundle_io.cc
@@ -23,6 +23,7 @@
 #include "math/vector.h"
 #include "mve/bundle_io.h"
 #include "mve/image_tools.h"
+#include "mve/depthmap.h"
 
 
 MVE_NAMESPACE_BEGIN
@@ -549,8 +550,15 @@ create_camera_info_from_params(CameraInfo& camera_info,
         // Pinhole: fx, fy, cx, cy
         float fx = params[0];
         float fy = params[1];
-        camera_info.flen = fx;
-        camera_info.paspect = fy / fx;
+        float dim_aspect = static_cast<float>(width) / height;
+        float pixel_aspect = fy / fx;
+        float img_aspect = dim_aspect * pixel_aspect;
+        if (img_aspect < 1.0f) {
+            camera_info.flen = fy / height;
+        } else {
+            camera_info.flen = fx / width;
+        }
+        camera_info.paspect = pixel_aspect;
         camera_info.ppoint[0] = params[2] / width;
         camera_info.ppoint[1] = params[3] / height;
     }
@@ -1130,45 +1138,6 @@ parse_colmap_depth_map(const std::string& path)
     return depth_img;
 }
 
-void
-convert_depth_map_semantics(mve::CameraInfo& mve_cam, int original_width,
-    int original_height, mve::FloatImage::Ptr depth_image)
-{
-    // The meaning of the depth map values in MVE and Colmap differs.
-    // In MVE the depth value is defined w.r.t. unit vectors.
-    // In Colmap the depth value is defined w.r.t. canonical vectors (i.e.
-    // the z-component is equal to 1).
-    size_t max_extent = std::max(original_width, original_height);
-    float fx = mve_cam.flen * max_extent;
-    float fy = fx * mve_cam.paspect;
-    float cx = mve_cam.ppoint[0] * original_width;
-    float cy = mve_cam.ppoint[1] * original_height;
-
-    size_t depth_map_width = depth_image->width();
-    size_t depth_map_height = depth_image->height();
-
-    float u_step_size = original_width / depth_map_width;
-    float v_step_size = original_height / depth_map_height;
-
-    for (int idx = 0; idx < depth_image->get_pixel_amount(); ++idx)
-    {
-        size_t u = idx % depth_map_width;
-        size_t v = idx / depth_map_width;
-        // No offset of 0.5 here, since Colmap depth values are defined w.r.t.
-        // shifted pixel values
-        float u_img = u_step_size * (float)u;
-        float v_img = v_step_size * (float)v;
-        // The cannoncial vectors are defined according to p.155 of
-        // "Multiple View Geometry" by Hartley and Zisserman
-        math::Vec3f canonical_ray{
-            (u_img - cx) / fx,
-            (v_img - cy) / fy,
-            1.0};
-        float norm = canonical_ray.norm();
-        depth_image->at(idx) = depth_image->at(idx) * norm;
-    }
-}
-
 mve::FloatImage::Ptr
 load_colmap_depth_map(int scale, mve::CameraInfo& mve_cam, int original_width,
     int original_height,
@@ -1182,8 +1151,11 @@ load_colmap_depth_map(int scale, mve::CameraInfo& mve_cam, int original_width,
     int depth_width = depth_image->width();
     int depth_height = depth_image->height();
 
-    convert_depth_map_semantics(mve_cam, original_width, original_height,
-        depth_image);
+    math::Matrix3f inv_calib;
+    mve_cam.fill_inverse_calibration(*inv_calib, original_width,
+        original_height);
+    mve::image::depthmap_convert_conventions<float>(depth_image, inv_calib, 
+        true);
 
     if (depth_width == original_width && depth_height == original_height)
     {


libs/mve/depthmap.h
```diff
@@ -167,16 +167,17 @@ MVE_IMAGE_NAMESPACE_BEGIN
 template <typename T>
 inline void
 depthmap_convert_conventions (typename Image<T>::Ptr dm,
-    math::Matrix3f const& invproj, bool to_mve)
+    math::Matrix3f const& invproj, bool to_mve, bool add_pixel_offset)
```
simonfuhrmann, line 170: I'm curious, why did you add this? This pixel offset is always necessary when converting from image coordinates (where the center of the top left pixel is (0,0)) to 3D image plane coordinates (where the center of the top left pixel is (0.5,0.5)).

SBCV: From the following comment in [colmap/blob/dev/src/base/reconstruction.cc](https://github.com/colmap/colmap/blob/dev/src/base/reconstruction.cc#L1036),

      // Bundler output assumes image coordinate system origin
      // in the lower left corner of the image with the center of
      // the lower left pixel being (0, 0). Our coordinate system
      // starts in the upper left corner with the center of the
      // upper left pixel being (0.5, 0.5).

I understand that the **image** coordinate system of Colmap also starts at (0.5,0.5). I made some experiments a few weeks ago, which "confirmed" that.

SBCV: Quick question: Does MVE offer some way to associate something like view_0006.mve with the original image name? Since the images are also reordered it is not easy to match the results with the original images

simonfuhrmann: Why is it relevant what convention Colmap uses internally for images? You're inputting images into MVE, and the pixels of that depth image correspond to the pixels of the color image. Now you're treating these images with MVE's convention to convert to a depth map in MVE's convention.

Unless you're telling me that that the depth map is shifted by 0.5 pixels compared to the color image, no change should be necessary here.

Each MVE view has a `meta.ini` file that has a `name` key. Typically you put the original image name (such as `IMG_123.JPG`) into that key. Does that help?

SBCV: The Colmap Fork I'm working with provides some matrices, which transform the reparameterized depth maps back into world coordinates. In the following, I'll use the corresponding point clouds as reference (shown in red).

Here, an overlay of the reference point cloud and the point cloud generated with offset=0.0 (green). The reason it does not completely match is the fact that the computations in the reparameterized space are numerically more stable. But one can see that the differences (the noise) appears to be random and non-systematic.

no_offset

Here, an overlay of the reference point cloud and the point cloud generated with offset=0.5 (green)
with_offset
 
On can clearly see that there is a systematic difference between the point clouds.

I think that makes sense, or am I missing something here?

simonfuhrmann: But hold on a second. The offset we're talking about here is solely used to compute the scaling factor for converting depth values. I would assume that this offset has a negligible effect on the depth actually.

If the effect was actually visible, wouldn't we see slightly smaller depth values on the left side of the image (in front of geometry), and slightly larger depth values on the right side of the image (behind geometry)? So wouldn't a systematic artifact manifest in a greenish image on the left half, and a reddish image on the right half?

I'm not sure that's the problem we're seeing here.

Then again, there is also a chance that the Colmap generated geometry is shifted when raising the points to 3D space?

simonfuhrmann: Suggestion. Let's submit your PR without this boolean first. Then let's discuss this issue in a separate PR.

simonfuhrmann: Do you maybe have to add/subtract 0.5 from the principal point (before normalizing it) when Colmaps image conventions are different?

SBCV: > Why is it relevant what convention Colmap uses internally for images? You're inputting images into MVE, and the pixels of that depth image correspond to the pixels of the color image. Now you're treating these images with MVE's convention to convert to a depth map in MVE's convention. Unless you're telling me that Colmap's depth map is shifted by 0.5 pixels compared to the color image, no change should be necessary here.

I understand it as follows:
Obviously, each value in a depth map (together with the camera parameters) represents a 3D point in camera / world coordinates. Thus,  the meaning of the actual value is completely depending on the the other parameters. Assuming that colmap defined the depth values with respect to 
`math::Vec3f px((float)x, (float)y, 1.0f); ` instead of `math::Vec3f px((float)x + 0.5, (float)y + 0.5, 1.0f);`, I need to use exactly this computation to derive the original point in camera/world coordinates. 

Because of the comment in `reconstruction.cc`, I expected (assuming that Colmap images and Colmap depth maps are aligned) that `math::Vec3f px((float)x, (float)y, 1.0f); ` is the correct computation

simonfuhrmann: Let me clarify what convention MVE uses for images, in case I didn't make that clear.

When accessing image (such as mve::ByteImage), `(0,0)` is the center of the first pixel. And when you use linear interpolation, for example, `(0,0)` will give you the first pixel's color (and not an interpolated color). `(0.5, 0)` will give you the interpolation between the first and second pixel. The horizontal index starts at `0` and ends at `width - 1`. This implies that the valid image region is in range `[-0.5, width - 0.5] x [-0.5, height - 0.5]`. I think this is pretty standard convention.

Now, when we're talking about the 3D space, and looking at the image plane in 3D space, that plane has dimensions `[0, width] x [0, height]`, and the upper left corner corresponds to `(-0.5, -0.5)` in image coordinates. And when you convert pixel `(x,y)` to a 3D point you'll get a coordinate `(x + 0.5, y + 0.5, flen)`.

I hope this makes sense.

SBCV: > But hold on a second. The offset we're talking about here is solely used to compute the scaling factor for converting depth values. I would assume that this offset has a negligible effect on the depth actually.

Why? With this offset the position of the 3D point changes. And MVE uses the distance to the point as depth, so the value will be different. No?

> If the effect was actually visible, wouldn't we see slightly smaller depth values on the left side of the image (in front of geometry), and slightly larger depth values on the right side of the image (behind geometry)? So wouldn't a systematic artifact manifest in a greenish image on the left half, and a reddish image on the right half?

Acutally, I don't dare to make any speculations about this, since the camera models are so crazy. See this intrinsic parameters as example: `fx, fy, skew, cx, cy: [4150290.730193433, 4082933.0617155796, -5788.048198890227, 46942.580589824625, 718815.7955793234]`
Note, that I remove the skew before processing the results with MVE.

> Suggestion. Let's submit your PR without this boolean first. 

Okay

SBCV: > Then let's discuss this issue in a separate PR.

Actually, I have no idea how we could further verify this. So what would you expect to be contained in this separate PR?
Since you are the maintainer of the repository, I think you should decide how we handle this ;)

SBCV: > >     But hold on a second. The offset we're talking about here is solely used to compute the scaling factor for converting depth values. I would assume that this offset has a negligible effect on the depth actually.
> 
> Why? With this offset the position of the 3D point changes. And MVE uses the distance to the point as depth, so the value will be different. No?

This referred to the scaling value

simonfuhrmann: In this code

```
  math::Vec3f px((float)x + 0.5f, (float)y + 0.5f, 1.0f);
  px = invproj * px;
  double len = px.norm();
  dm->at(pos) *= (to_mve ? len : 1.0 / len);
```

the position of `px` does change with the offset, but it's really only used to scale the depth value (`px.norm()`), i.e., changes the final point cloud position in viewing direction, not orthogonal to it. Have you tried adding/subtracting 0.5 from the principal point before normalizing it (as per my previous comment). Maybe this fixes this issue you're seeing?

simonfuhrmann: This is to say, if the `0.5` offset was the problem, there would be more places you'd have to fix.
Most notably [this one](https://github.com/simonfuhrmann/mve/blob/master/libs/mve/depthmap.cc#L150).

SBCV: > Have you tried adding/subtracting 0.5 from the principal point before normalizing it (as per my previous comment). Maybe this fixes this issue you're seeing

No, but this could / should have the same effect. 

> This is to say, if the 0.5 offset was the problem, there would be more places you'd have to fix.

I'm afraid that's the case. 

Oh boy this importer is killing me ... :D

simonfuhrmann: I feel you man. These two things are among the hardest in human history: Converting between conventions, and rotations. Literally the hardest thing in the world is converting rotation conventions.

SBCV:
Offset reverted and squashed commit

SBCV:
I've conducted some more experiments and I think you are right with 

> I'm not sure that's the problem we're seeing here.

The previous results were created with a python script, I already had to convert the depth maps to world points. 
When using MVE's  `scene2pset` with a single `id` to create the world points corresponding to the depth map of `id`, I can not recreate this effect. (Adjusting the principal point (with -0.5f) did actually produce worse results)

In addition, I run some texturing experiments with the Sceaux dataset that showed that adjusting the principal point (with -0.5f) leads to a mesh with untextured faces. 

Thus, I think the master branch is already in a good shape. 

Sorry for the noise and thank you for merging.

simonfuhrmann:
No worries, I'm glad it works. Just to reiterate on the principal point though. When you subtract 0.5 from it, it has to be before normalization. For example, this

`camera_info.ppoint[0] = params[2] / width;`

Will become

`camera_info.ppoint[0] = (params[2] - 0.5) / width;`

SBCV:
> The previous results were created with a python script, I already had to convert the depth maps to world points.
> When using MVE's scene2pset with a single id to create the world points corresponding to the depth map of id, I can not recreate this effect. (Adjusting the principal point (with -0.5f) did actually produce worse results)

I left out a lot of information in the paragraph above to keep it short, but reading it again, makes me realize that this paragraph is probably misleading. Here some more background information so you can fully understand the previous / following images:
The Colmap Fork for the reconstruction of satelite images
- contains camera models with skew parameters
- produces depth maps in reparameterized space (the values are way smaller than the actual depth map values)

Let us denote such a depth map as `depthmap_{sr}`, i.e. the depth maps are skewed and contain reparameterized depth values. In order to use the depth maps we need to convert `depthmap_{sr}` to depth maps with real world depth values, in the following denoted as  `depthmap_{s}` (the depth map is still skewed). The following result (already shown in a post 2 days ago)
image
is the point cloud in world space computed from  `depthmap_{s}`. Reminder: The red point cloud is a reference point cloud directly computed with the colmap fork.

In order to use the depth maps in MVE, we need further to remove the skew factor from the depth maps  `depthmap_{s}` (and of course we also need to remove the skew from the corresponding camera models and images). Let us denote such a skew-free depth map as  `depthmap_{MVE}`.

Now, regarding 
> For example, this
> camera_info.ppoint[0] = params[2] / width;
> Will become
> camera_info.ppoint[0] = (params[2] - 0.5) / width;

Here are some result with the point cloud generated by `scene2pset` (using such a skew-free depth map  `depthmap_{MVE}`) and the "reference point cloud". Please ignore the pattern in the following results, I think that is some artifact of the skew removal - the maps must be warped at this step. I need to further investigate this.

Result without principal point adjustment (`camera_info.ppoint[0] = params[2] / width;`). As you can see sometimes the reference is in front and vice versa.

no_pp_offset

Result with subtraction (`camera_info.ppoint[0] = (params[2] - 0.5) / width;`). As you can see, the reference is largely in the background.

negative_pp_offset

Result with addition (`camera_info.ppoint[0] = (params[2] + 0.5) / width;`). As you can see, the reference is largely in the foreground.

positive_pp_offset

Thus, `camera_info.ppoint[0] = (params[2] - 0.5) / width;` as well as  `camera_info.ppoint[0] = (params[2] + 0.5) / width;` introduce an additional shift, which lead to worse results. 

> In addition, I run some texturing experiments with the Sceaux dataset that showed that adjusting the principal point (with -0.5f) leads to a mesh with untextured faces.

Here, a result with 
`camera_info.ppoint[0] = params[2] / width;`
no_pp_offset

and a result with 
`camera_info.ppoint[0] = (params[2] - 0.5) / width;`

with_pp_offset

As you can see the result with the offset introduces untextured (white/grey) faces, which further support that `camera_info.ppoint[0] = (params[2] - 0.5) / width;` (and `camera_info.ppoint[1] = (params[3] - 0.5) / height;`) is correct.

Looking at these results, I think we should not touch the principal point.

SBCV:
So, by using `scene2pset` with `depthmap_{MVE}`, I could not reproduce the effect shown in the comparison using `depthmap_{s}`. (But maybe the reason for this is that the artifacts of the warping have are stronger effect than `math::Vec3f px((float)x + 0.5f, (float)y + 0.5f, 1.0f)`). So, potentially it could be still there. But like you pointed out earlier

> Unless you're telling me that Colmap's depth map is shifted by 0.5 pixels compared to the color image, no change should be necessary here.

that would mean that there is a shift between the image and the depth map, which sounds really unreasonable to me. 
So therefore, I thought we should keep the master branch as it is. 

However, looking again at the results in `math::Vec3f px((float)x + 0.5f, (float)y + 0.5f, 1.0f)` makes me just realize how perfectly (!) the point clouds match with this additional shift. 

(see results below - copied from a previous post)

image

and 
image

But then again, that would mean that there is this (non-intuitive) shift ...

PR #519, opened 2020-09-09
improve name key

SBCV:


SBCV:
Oh wait, don't merge now

SBCV:
Okay, now in clean state

PR #535, opened 2021-02-04
Fix travis build issue

keineahnung2345:
I've noticed that [recent travis builds](https://travis-ci.org/github/simonfuhrmann/mve/builds) failed for a reason that has nothing to do with the code.

After some searching, I've found that [cmake should be run in googletest's home directory, not googletest/googletest](https://github.com/google/googletest/issues/2882#issuecomment-753471115).

So this PR revise `GTEST_PATH` and update the include directory in `tests/Makefile` accordingly.

You can see that these changes make travis build successfully: https://travis-ci.org/github/keineahnung2345/mve/jobs/757523169.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,7 +23,6 @@ before_script:
    - wget https://github.com/google/googletest/archive/master.zip -O /tmp/gtest.zip
    - unzip /tmp/gtest.zip
    - export GTEST_PATH=$PWD/googletest-master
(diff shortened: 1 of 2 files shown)


simonfuhrmann:
Hi there, would be mind squashing all the commits in to one?
BTW, this fix is really appreciated. gtest breaks once in a while because they are making directory changes or changes to the build system, it's kinda annoying...

keineahnung2345:
> Hi there, would be mind squashing all the commits in to one?

Please go ahead.

> BTW, this fix is really appreciated. gtest breaks once in a while because they are making directory changes or changes to the build system, it's kinda annoying...

I agree, they should update their doc as soon as possible.

PR #545, opened 2021-05-19
Check extensions before loading images/headers

pierotofy:
Hello :hand: !

This PR proposes to add checking to file extensions before attempting to load a file. Aside from the modest speed improvement, this solves a deeper issue of leaking file descriptors with JPG files (which happens if you load TIFFs or ppm or mvei, due to the order in which file formats are tried).

The root cause of the leaks is that attempting to read an invalid JPG file (e.g. a TIFF file) will trigger the error callback in https://github.com/simonfuhrmann/mve/blob/master/libs/mve/image_io.cc#L467 `jpg_error_handler`, but because the function is a library call, the util::Exception triggered in the callback is not caught by the catch-all function block, and is instead caught further up the stack, causing `fp` to never be closed.

This doesn't explicitly fix the leak (I'm unsure of a good way to handle this gracefully), but the extension check is a nice way to avoid trying to open obvious files with the wrong reader (and mitigates the leak).﻿

### DIFF
--- a/libs/mve/image_io.cc
+++ b/libs/mve/image_io.cc
@@ -56,33 +56,49 @@ MVE_IMAGE_NAMESPACE_BEGIN
 ByteImage::Ptr
 load_file (std::string const& filename)
 {
+    using namespace util::string;
+    std::string fext4 = lowercase(right(filename, 4));
+    std::string fext5 = lowercase(right(filename, 5));
+
     try
     {
 #ifndef MVE_NO_PNG_SUPPORT
-        try
-        { return load_png_file(filename); }
-        catch (util::FileException& e) { throw; }
-        catch (util::Exception& e) {}
+        if (fext4 == ".png") {
+            try
+            { return load_png_file(filename); }
+            catch (util::FileException& e) { throw; }
+            catch (util::Exception& e) {}
+        }
 #endif
 
 #ifndef MVE_NO_JPEG_SUPPORT
-        try
-        { return load_jpg_file(filename); }
-        catch (util::FileException& e) { throw; }
-        catch (util::Exception& e) {}
+        if (fext4 == ".jpg" || fext5 == ".jpeg") {
+            try
+            {
+                return load_jpg_file(filename);
+            }
+            catch (util::FileException& e) { throw; }
+            catch (util::Exception& e) {}
+        }
 #endif
 
 #ifndef MVE_NO_TIFF_SUPPORT
-        try
-        { return load_tiff_file(filename); }
-        catch (util::FileException& e) { throw; }
-        catch (util::Exception& e) {}
+        if (fext4 == ".tif" || fext5 == ".tiff") {
+            try
+            { return load_tiff_file(filename); }
+            catch (util::FileException& e) { throw; }
+            catch (util::Exception& e) {}
+        }
 #endif
 
-        try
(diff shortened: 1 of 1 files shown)


simonfuhrmann:
Hi Piero, let me take a look at the underlying issue. It would be nicer to properly not leak the file descriptors. While it is probably a fair assumption that file extensions should be correct, sometimes file extensions can be weird ".jpg, .jpeg., .JpEG, .tif, .tiff". While it's easy enough to handle those cases, sometimes they are straight up wrong, and the "header peaking" mechanic in MVE would still lead to a proper loading. So in a way, we would introduce a regression.

simonfuhrmann:
Piero, are you sure the JPEG reading is leaking file descriptors? I just put some debugging code in the function and got this while attempting to load a PNG file with `load_jpg_file()`:

```
std::fopen
JPG error handler
std::fclose catch
terminate called after throwing an instance of 'util::Exception'
  what():  JPEG format not recognized
```

The flow is as follows. It opens the file, sets the error handler, does some loading, JPEG realizes it's not a JPEG, calls the error handler, and the error handler throws. Since all of the JPEG loading happens in the try/catch block, the catch block closes the file descriptor.

Let me know what you thing.

pierotofy:
It might have to do with the environment/compiler. I should have added this in the description, but I've found this problem while running MVE on Windows (compiled with Visual Studio 2019). I think this boils down to the fact that throwing C++ exceptions from a C callback is an undefined behavior (some details are concisely expressed in https://stackoverflow.com/questions/10956062/throwing-an-exception-in-c-in-a-c-callback-possibly-crossing-over-dynamic-lib ).

In the debugging session I did on Windows, the catch block would get skipped (and be caught at the upper level). I should note this would also be followed by a "stack corruption" error code at exit, so not sure if something funkier was happening that was happening the catch block to be skipped.

The extension casing cases (`JPeG` vs `jpeg`, etc.) should be handled by the call to `lowercase`.

I thought about fixing the root cause of the leak, however, it's not exactly straightforward due to the nature of libjpeg (header read errors are reported always via callback), but since the code could be multi-threaded, there would be a need to track load-state of images in a global context and I'd rather not go there. The extension check seems like a good solution because it covers most use cases.

Fully understand the concern about the regression however, if you don't want to change the behavior I'd recommend not to include these changes. Just thought I'd share the code if there was interest.

simonfuhrmann:
Ok thanks for the link. I understand a bit more about the issue now. Have to tried using this?
https://stackoverflow.com/questions/15845681/can-c-functions-marked-as-extern-c-throw/24242183#24242183
Ideally, a compiler flag would just fix this issue...

pierotofy:
I haven't tried to use that flag, but it sounds like it may work too.

pierotofy:
I'll close this PR since there's probably a way to avoid this issue with a compiler flag. It will avoid the possibility of regressions. :+1:  Thanks for the review!

simonfuhrmann:
Piero, my intention was not to create a barrier for fixing this issue. There are still ways we can make this work without compiler flags (e.g., I think this `longjmp` was mentioned as an alternative to exceptions). Have you tried the compiler flag, and does ti work?

pierotofy:
I haven't tried; the workaround of using extension checks currently works well for the use case I was fixing. I'm not sure when I'll have time to test due to priorities.

PR #577, opened 2024-09-05
UMVE: Raise C++ version to C++14

andre-schulz:
The C++ version of MVE was raised to C++14 in a previous commit and UMVE was missed.


simonfuhrmann:
Thaaaaanks!!!1

PR #582, opened 2025-11-03
switch to c++17

simogasp:
Switch to c++17 standard as recent versions of GTest require it. The CI on macos was failing because homebrew pulls such versions.

Incidentally, 
- added permissions to the workflow
- set a monthly schedule for automatically triggering the workflow

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -25,13 +25,10 @@ jobs:
 
       matrix:
         platform:
-        - { name: "Ubuntu 20.04, GCC, x86_64",   os: ubuntu-20.04, cpp_compiler: g++,     qmake_spec: linux-g++ }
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -5,6 +5,15 @@ on:
     branches: [ "master" ]
   pull_request:
     branches: [ "master" ]
+  workflow_dispatch:
+  schedule:
+    # every 1st day of month at 00:00 UTC
+    - cron: "0 0 1 * *"
+
+permissions:
```
simonfuhrmann: Can you explain this section to me? Why does this workflow need non-default permissions? What does `checks: write` and `pull-requests: write` mean?

simonfuhrmann:
Thanks.

